PR #2, opened 2025-09-09
Add small/bigger or equal operators

rTreutlein:


patham9 (approved):
Looks great, thank you!

PR #4, opened 2025-09-17
Lightweight Janus-based Python integration

patham9:
# Python bindings: adopt the minimalist Janus wrapper

**Summary**  
This PR replaces the Hyperon-Experimental-compatible dynamic-dispatch approach in the `janus2` branch with a **thin, explicit wrapper** over Janus.  It won the competition regarding **clarity, elegance, and low overhead**.

---

## Why this approach?
- **No runtime guessing.** We don’t pass Python callables around and then repeatedly test if something is callable.  
- **No `eval` and dict building for each call.** Direct calls via `py_call(Mod:Fun(...))` and `Obj:Method(...)`.  
- **Minimalistic, yet generic.** Supports `.method`, `mod.fun`, `fun`, and `getattr` cases uniformly.

---

## Advantages
- **Performance:** Less runtime dispatch and zero `eval` and dict building → smaller per-call overhead, especially in tight loops.  
- **Elegance:** Each py-call invocation, e.g. `'py-call'(['torch.matmul', A, B], D)` maps directly to Python’s own call rules (`.method`, `mod.fun`, `getattr`) with no extra indirection.

---

## Disadvantages
- **Compatibility:** How PeTTa calls Python functions is not directly compatible with Hyperon-Experimental, however since Janus for Prolog is mature and already existed, a thin wrapper over it is favourable compared to a complex wrapper for the sake of API compatibility with Hyperon-Experimental.

---

## Examples
- Minimalist wrapper (**this PR**):  
  - <https://github.com/patham9/PeTTa/blob/janus/examples/python.metta>  
  - <https://github.com/patham9/PeTTa/blob/janus/examples/torch.metta>

- Previous approach (**rejected, kept for reference**):  
  - <https://github.com/patham9/PeTTa/blob/janus2/examples/torch.metta>

---

## Notes on `janus2` (rejected)
- Relies on dynamic dispatch (`maybe_call/3`) → repeated callable checks in general MeTTa code.  
- Treats Python functions like first-class MeTTa functors → **heavier Prolog** and slower hot paths.  
- Re-introduces “explicit conversion” helpers already unnecessary with Janus’ direct Prolog↔Python mapping.


src/parser.pl
```diff
@@ -19,7 +19,7 @@
 var_symbol(V,E0,E) --> "$", token(Cs), { atom_chars(N, Cs), ( memberchk(N-V0, E0) -> V=V0, E=E0 ; V=_, E=[N-V|E0] ) }.
 
 %Atoms are just tokens:
-atom_symbol(A) --> token(Cs), { atom_codes(A0,Cs), downcase_atom(A0,A) }.
+atom_symbol(A) --> token([C0|Rest]), { code_type(C0, to_upper(L0)), atom_codes(A, [L0|Rest]) }.
```
patham9, line 22: Only first letter is made lowercase. This is a good tradeoff as that way Python interop is not restricted via module.ClassName for instance while when a user is writing MeTTa code the user does not need to care about upper or lowercase in naming.

PR #6, opened 2025-09-24
Added more functions to stdlib

rTreutlein:
I just added some stuf i needed. We probably want a more structed effort to expand the stdlib. Also not sure all the functions i have in here should be in the stdlib.

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -82,19 +82,36 @@ list_to_set_helper(Pred, [H|T], Acc, Set) :-
     ;   list_to_set_helper(Pred, T, [H|Acc], Set)
     ).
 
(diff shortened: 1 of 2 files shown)


patham9 (approved):
Looks great, thank you very much! I wonder why alpha equality is registered twice, is that the case in H-E too? I never tried that in H-E.

rTreutlein:
> Looks great, thank you very much! I wonder why alpha equality is registered twice, is that the case in H-E too? I never tried that in H-E.

Well i couldn't decide which one to use. =alpha i have cause that's what it's called in HE but i actually prefer =@= from prolog so i just added both.

PR #13, opened 2025-09-26
Add a Python Wrapper Library

rTreutlein:


patham9 (changes requested):
Thank you for this attempt. Hm to my understanding pyswip is an outdated wrapper that has way more overhead than Janus. Since PeTTa already invokes Janus by itself it might be better to switch to Janus.

rTreutlein:
I changed the way run works. Had to give it an extra argument. If you think thats an okay solution then you can merge this now.

rTreutlein:
True if they retract themself that would be nice aswell.

patham9:
What is this file doing? 
image

PR #21, opened 2025-10-08
Move the major, most specific, premise of modus ponens first

ngeiswei:
This has the effect of considerably speeding up the search because the backward chainers expends premises in their order of appearances.  As a result all proves can be discovered instead of merely checked.


patham9 (approved):
Awesome, even the most difficult proof? :)
Thank you very much!

PR #26, opened 2025-10-13
Fix parsing 

rTreutlein:
previously 2025_12_12 was translated to [2025, '_12_12'] instead of '2025_12_12'


rTreutlein:
replaced by https://github.com/patham9/PeTTa/pull/28

PR #31, opened 2025-10-18
Add unique with test

rTreutlein:


patham9 (changes requested):
unique in H-E is not operating on explicit lists:
`!(test (unique (1 1 $a $a)) (1 $a))`
We should probably make it consistent with H-E and use list_to_set for explicit lists.

(E.g. if we have new stdlib functions we can have them ofc, while if they have the same name as in H-E they should ideally also behave like in H-E)

PR #35, opened 2025-10-20
Fix of few prolog function and addition of math functions implementation 

abrham17:
#### Union-atom and Union
- **Issue**: The `union` function converts lists to sets, removing duplicates, unlike the intended behavior of `union-atom` and `union` which should combine elements without removing duplicates. It also doesn't accept non-deterministic input.
- **Suggestion**: Rename the compiler's builtin `concat` function to `union-atom` and fix the issue, as `concat` works similarly to `union-atom`.

#### Intersection-atom and Intersection
- **Issue**: PeTTa's `intersection` includes the largest frequency of elements in both expressions, while Metta's `intersection-atom` keeps the smallest.
- **Suggestion**: Create a wrapper `intersection-atom` function accepting 2 arguments with default parameters, but this won't fix the frequency issue.

#### Size-atom
- **Issue**: Prolog's `length` function is used instead of `size-atom`.
- **Suggestion**: Replace `length` with a custom `size-atom` function and register it.

#### Cons-atom
- **Issue**: Uses `cons` instead of `cons-atom`.
- **Suggestion**: Rename to `cons-atom` and register it in `Metta.pl`.

#### Subtraction-atom
- **Issue**: Uses `subtract` instead of `subtraction-atom`.
- **Suggestion**: Write a `subtraction-atom` wrapper function using `subtract` and register it.

#### Metta's Maths builtin functions
- **Sqrt-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `sqrt` function with `sqrt-math` and register it.
- **Abs-math**: Not implemented.
  - **Suggestion**: Implement with a custom absolute value function and register `abs-math`.
- **Log-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `log` function with `log-math` and register it.
- **Trunc-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `truncate` with `trunc-math` and register it.
- **Ceil-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `ceiling` with `ceil-math` and register it.
- **Floor-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `floor` with `floor-math` and register it.
- **Round-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `round` with `round-math` and register it.
- **Cos-math, tan-math, sin-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `cos`, `tan`, and `sin` with respective functions and register them.
- **Asin-math, acos-math, atan-math**: Not implemented.
  - **Suggestion**: Wrap Prolog's `asin`, `acos`, and `atan` with respective functions and register them.
- **Isnan-math and isinf-math**: Not implemented.
  - **Suggestion**: Implement with custom logic for NaN and infinity checks and register them.
- **Min-atom and max-atom**: Not implemented.
  - **No suggestion provided.**
- **Random-int and Random-math**: Not implemented.
  - **Suggestion**: Implement `random-int` and `random-float` without space parameter and register them.


patham9 (changes requested):
Thank you for reporting your findings and improvement suggestions via this PR! This will take some review time, and some of the naming conventions we might want to discuss. Also whether we prefer Prolog naming of predicates that are already functions with last-arg-output over rolling our own or renaming it. Also some naming in H-E is annoying, e.g. -atom suffix.

patham9:
I agree @abrham17 . Since last time some stdlib functions have been added: https://github.com/patham9/PeTTa/blob/main/src/metta.pl If you see some missing from your PR, can you please open a new PR so we can merge your additions which are not yet there? size-atom for instance. Regarding intersection using alpha-equality: H-E also does not seem to do that, or do newer versions do that?

PR #37, opened 2025-10-22
Fix: standard name issue and add general map-atom function

tinsaeabreham45:
This PR updates the names of car, cdr, cons, and decons primitives to include the -atom suffix (e.g., decons-atom). This resolves a naming consistency issue and adds the general map-atom function.
History Note: This branch was rebased onto main to ensure a clean, linear history.

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -59,10 +59,21 @@ empty(_) :- fail.
 .(A, B, R) :- R=[A|B].
 'car-atom'([H|_], H).
 'cdr-atom'([_|T], T).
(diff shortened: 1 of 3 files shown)


patham9 (changes requested):
At current stage this PR has a problematic merge solution deleting key components like eval. I highly recommend to change only a single thing per PR, else it is hard to cherry-pick the good things of each PR even though I try my best to do so even when a PR as a whole cannot be merged. Also the more you change at once, the more likely you will end up with merge conflicts during development.

PR #40, opened 2025-10-24
Comprehensive update to == operator: supports numbers, atoms, lists, and nested terms

yabuz87:
## 🧩 Description (updated)
This PR enhances the `==` function to support **comprehensive equality checking** across multiple MeTTa data types.

Previously, `(== 2 2.0)` returned `false` because Prolog's `==` operator compares **terms** rather than **values**, and other types such as atoms, lists, or nested structures were not supported.

This update introduces:

1. **Numeric equality:** Correct comparison between integers and floats (e.g., `(== 1 1.0)` → `true`).  
2. **Atom and string equality:** Exact matching of atoms and strings.  
3. **Boolean equality:** `(== true true)` and `(== false false)` now work correctly.  
4. **List equality:** Recursive comparison for lists and nested lists.  
5. **Compound term equality:** Compares functor and arguments for structured terms.  
6. **New test coverage:** Added multiple test cases for all supported types in the `examples` directory.

---

### 🧪 Example

**Before:**
```metta
(== 2 2.0)
; Output: false

(== (1 2 3) (1 2 3))
; Output: error

(== hello hello)
; Output: error
```
**now:**
```metta
(== 2 2.0)
; Output: true

(== (1 2 3) (1 2 3))
; Output: true

(== hello hello)
; Output: true
```

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -21,50 +21,91 @@ chain(Eval, Var, After, Out) :- let(Var, Eval, After, Out).
 '>'(A,B,R)  :- (A>B -> R=true ; R=false).
 '=='(A,B,R) :- (eq_comprehensive(A,B) -> R=true ; R=false).
 
-% Comprehensive equality checking for == operator
+% Optimized comprehensive equality checking for == operator
 % Handles int/float equality, atoms, lists, and nested structures
+% Performance optimizations: identity check, fast paths, normalized numeric comparison
+
 eq_comprehensive(A, B) :- 
-    % Handle numeric equality (int and float with equal magnitude)
+    % Fast path: identical objects
+    A == B, !.
+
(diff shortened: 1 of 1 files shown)


patham9 (changes requested):
Thank you! It is on the right track but not the solution as it also has to work for atoms, lists thereof etc.
image

yabuz87:
could you review it again please ?

patham9 (changes requested):
That change is well-executed and achieves the intended semantics. However, recursive equality checking is costly, so we should benchmark it comprehensively before merging. If performance becomes an issue (and I would be surprised if not), a normalization step combined with built-in equality operators might be a better approach. In worst-case leaving ints and floats separate is also not a big deal.

PR #42, opened 2025-10-28
intersection-atom , size-atom implementation fix

abrham17:
**Intersection-atom** 

**Issue**: PeTTa's intersection includes the largest frequency of elements in both expressions, while Metta's intersection-atom keeps the smallest.
**Suggestion**: Use select  **select** builtin prolog function in the intersection function definition. This will fix the frequency issue.

  **Size-atom**

 **Issue**: Prolog's length function is used instead of size-atom.
 **Suggestion**: Replace length with a custom size-atom function and register it.


patham9 (changes requested):
There is a point to rename the more flexible member_with_pred versions consistently, but to remove them is not an option as they are more powerful than the H-E variant. intersection is already supported as in H-E,  and it does not operate on lists in H-E either, only intersection-atom does, does that suffer from the issue you described? I'm comfortable to add 'size-atom' as-is though. And happy to look at test cases as it is always hard to see what changes are about without supporting tests they enable. Thank you for your effort!

patham9:
Added size-atom in the meanwhile: https://github.com/patham9/PeTTa/blob/main/src/metta.pl#L90

PR #43, opened 2025-10-28
Add nondeterminism predicates for union, unique, and intersection

abrham17:
**Issue**
Non-deterministic set operation in metta have not been implemented in PeTTa.

**suggestion**
Implemented non-deterministic set operations using member and findall functions.

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -152,7 +152,17 @@ assert(Goal, true) :- ( call(Goal) -> true
                                       format("Assertion failed: ~w~n", [RG]),
                                       halt(1) ).
 
-%%
+%%% Nondeterminism %%%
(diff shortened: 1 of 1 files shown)


patham9:
"Non-deterministic set operation in metta have not been implemented in PeTTa."

They have how else do the test cases pass?

Tests:
https://github.com/patham9/PeTTa/blob/main/examples/streamops.metta
https://github.com/patham9/PeTTa/blob/main/examples/multiset_operations.metta

Implementation:
Multisets: https://github.com/patham9/PeTTa/blob/main/src/metta.pl#L100
Stream ops: https://github.com/patham9/PeTTa/blob/main/src/translator.pl#L37

Your implementation is incorrect since union/intersection in H-E operates on superpositions and not on lists.

PR #44, opened 2025-10-28
feat: Implement MeTTa-compatible error handling system with Error constructor

yabuz87:
- Add Error constructor supporting MeTTa's error representation format
- Provide error utilities for building robust MeTTaLog applications
- Fix function export issues to make error handling available to users
- Supports patterns like: (Error (add "a" 2) BadType) per MeTTa spec

````
(= (divide  $d $a)
(if (== $a 0)
  (Error ($a) (divsion by zero error))
    (/ $d $a)
))

!(divide 4 2) ; returns 2 
!(divide 4 0) ; returns  (Error (0) (divsion by zero error))  as previouse MeTTa and MeTTalog version

```


patham9 (changes requested):
Thanks for your efforts. But looks like some weird H-E functionality that would better fit in lib_he.metta instead of PeTTa itself? And is-error I did not see in H-E hm.

yabuz87:
Thanks for the feedback! The he_error file already provides sufficient error-handling functions, so this one isn’t needed.

PR #45, opened 2025-10-29
feat: Add random number generation utilities

yabuz87:
- Implement random-int for integer generation within specified intervals (inclusive)
- Add random-float for floating-point number generation

```
!(random-int min max)
!(random-float )
```


patham9 (approved):
Cool, thank you! :)

PR #46, opened 2025-10-29
Add 2 pass compilation and remove RunArg

rTreutlein:


patham9 (changes requested):
Great effort! Remaining issues: Formatting is not yet consistent with the rest of the project and `run` predicates would need to be removed entirely for it to become elegant.

PR #49, opened 2025-10-30
Factour out let/let*

rTreutlein:


patham9 (approved):
I LOVE this PR, so well-executed, elegant, perfect! Danke sehr Roman!

patham9:
Also 18 lines added, 24 removed, wonderful!
Any ideas about chain? If we can remove the let invocation there too then 5 lines, the runtime let/let* can be removed from metta.pl altogether.

patham9:
No worries about chain, I solved it already (will commit in few min)

PR #50, opened 2025-10-31
Remove tests requiring extra dependencies

rTreutlein:


rTreutlein:
Now we might also want to add those dependencies to our test env but for now that is the easiest fix.

PR #53, opened 2025-11-04
Add currying/partial application + lambdas

rTreutlein:
Also map/foldl/filter-atom all now use reduce to make currying work

The Lambad currently looks like in haskell which i like but we should probably change that.

### DIFF
--- a/examples/lambda.metta
+++ b/examples/lambda.metta
@@ -11,8 +11,8 @@
 !(test (applyL1) 3)
 !(test (applyL2) 9)
 
-!(test (maplist (\ ($a) -> (+ 1 $a)) (1 2 3))
+!(test (maplist (|-> ($a) (+ 1 $a)) (1 2 3))
        (2 3 4))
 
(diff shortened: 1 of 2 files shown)


patham9 (changes requested):
Nice attempt already, thank you!
Hm probably better to be prefix rather than infix?
E.g. 
`(\ ($a) -> (+ 1 $a))`
becomes: (math notation)
`(|-> ($a) (+ 1 $a))`
or (odd Haskell-ish style)
`(\-> ($a) (+ 1 $a))`

which is more consistent with the rest and also easier to write and read.

patham9:
Wonderful. looks close to finish line!
Last question: Why is partial separate from lambda? Wouldn't it be better it would be |-> all along so one can use curried functions in HOL functions?

rTreutlein:
Ah you mean given (= (f $a $b) $body) and (f $a) => (|-> ($b) ($f $a $b))
Didn't think of that but not sure it would work like that. Would have to try.

PR #55, opened 2025-11-06
Python Tests

rTreutlein:


patham9 (changes requested):
Looks good, but files like uv.lock should probably be excluded? Or is it fine?

PR #58, opened 2025-11-07
v1.0: Compile higher order predicates to specialized versions on call

rTreutlein:
It works but the code was generated by GPT-5 so there is likely a lot of cleanup we need to do before merging it. 

All current tests pass but we should also try for a bit to breaks this / add new tests.

### DIFF
--- a/examples/holbenchmark.metta
+++ b/examples/holbenchmark.metta
@@ -1,0 +1,34 @@
+(= (map-flat $f ()) ())
+(= (map-flat $f (cons $x $xs)) (cons ($f $x) (map-flat $f $xs)))
+
+(= (range $n)
+   (if (== $n 0) ()
+       (cons $n (range (- $n 1)))))
+
+!(test (let $temp (map-flat (+ 1) (range 1000000)) (length $temp)) 1000000)
+
+
+(= (fold-nested $f $init ()) $init)
+(= (fold-nested $f $init (cons $x $xs))
+      (if (is-expr $x)
+        (fold-nested $f (fold-nested $f $init $x) $xs)
+        (fold-nested $f ($f $init $x) $xs)))
+
+(= (deep-nest $n)
+   (if (== $n 0) ()
+       (cons (range 50) (deep-nest (- $n 1)))))
+
+!(test (fold-nested + 0 (deep-nest 20000)) 25500000)
+
+
+(= (apply-many $f $n $x)
+   (if (== $n 0) $x
+       (apply-many $f (- $n 1) ($f $x))))
+
+!(test (apply-many (+ 1) 100000 0) 100000)
+
+(= (poly $f $n)
+   (if (== $n 0) 0
+       (+ ($f $n) (poly $f (- $n 1)))))
+
+!(test (poly (+ 1) 1000000) 500001500000)


rTreutlein:
Alright i think this is ready to merge now.

patham9 (changes requested):
We will need to find a solution to allow function removal by tracking specializations. If a function is removed, all its specialization should vanish too. I don't think we need to re-compile functions a function is used in though, that complexity is usually not needed, also since users can still remove and add a function again if desired without introducing the unnecessary complexity. https://github.com/patham9/PeTTa/blob/main/examples/selfprog.metta

patham9:
Another point: all setup_call_cleanup and nb calls should go away. The clean way in Swi-Prolog to get around explicit argument passing is to use b_setval and b_getval which respects the call stack and unwinds correctly on backtracking.

src/translator.pl
```diff
@@ -304,13 +315,15 @@
 %Generate actual function call or partial if arity not complete:
 build_call_or_partial(Fun, AVs, Out, Inner, Extra, Goals) :- length(AVs, N),
                                                              Arity is N + 1,
-                                                             ( ( ( current_predicate(Fun/Arity) ; catch(arity(Fun, Arity), _, fail) ),
```
patham9: All as before except it attempts to specialize first.

src/translator.pl
```diff
@@ -216,13 +230,10 @@
                                            append(FreeVars, Args, FullArgs),
                                            % compile clause with all bound + free vars
                                            translate_clause([=, [F|FullArgs], Body], Clause),
-                                           ( silent(true) -> true ; format("\e[33m--> lambda clause -->~n\e[32m", []),
```
patham9: New clause printing mechanism which is called also from specializer

examples/specializecyclic.metta
```diff
@@ -0,0 +1,15 @@
+(= (f1 $f $a) (if (< $a 0)
```
patham9: Tests whether it works for cyclic dependencies while noticing whether the function is used or not used.

patham9:
First benchmark 2x speed improvement: https://github.com/patham9/PeTTa/blob/main/examples/patrick_iterate_quad.metta
Also translator uses b_setval for current.

patham9:
LAST MISSING PIECE **Remove all specializations too when parent function is removed. [TODO]**
Then the testcase will pass and the PR be ready for final review and merging.

patham9:
Yay it works, done. Except for testing that was the last missing piece for v1.0.

patham9 (dismissed):
done

src/specializer.pl
```diff
@@ -0,0 +1,107 @@
+:- dynamic ho_specialization/2.
+
+%Maybe specializes HV(AVs) if not already ongoing, and if specialization fails, nothing changes and specneeded is restored:
+maybe_specialize_call(HV, AVs, Out, Goal) :- setup_call_cleanup( (catch(nb_getval(specneeded,Prev),_,Prev = []), nb_setval(specneeded,false)),
+                                                                 specialize_call(HV, AVs, Out, Goal),
+                                                                 (Prev == true -> nb_setval(specneeded,Prev)) ).
```
patham9: specneeded acts as a latch between recursive translations. When any sub-function has the function at head position even when the parent/grandparent/kth-order-grandparent does not use it by itself, it means the k-order-parent needs specialization even when just to pass the special function to the child-chain. And in order to latch, we restore the previous value always even when we enter each specialization with specneeded=false.

src/specializer.pl
```diff
@@ -0,0 +1,107 @@
...
+eligible_var_pairs([], [], _, []).
+eligible_var_pairs([Var|Vars], [Copy|Copies], BodyExpr, HoVars) :- ( specializable_arg(Copy), (var_use_check(head, Var, BodyExpr) ; var_use_check(ho, Var, BodyExpr))
+                                                                     -> Var = Copy,
+                                                                        HoVars = [Var|RestHoVars]
+                                                                      ; HoVars = RestHoVars ),
+                                                                   eligible_var_pairs(Vars, Copies, BodyExpr, RestHoVars).
+
+%If Var appears at list head it means function call, meaning specialization is needed, and detect when used as HOL arg
+var_use_check(head, Var, [Head|_]) :- Var == Head,
+                                      nb_setval(specneeded, true).
```
patham9: That's the meat, a passed function at head position of a S-expression, indicating function call. Only when this condition is happening will specialization be triggered.

PR #59, opened 2025-11-07
Add chess game to petta repo

jazzbox35:
Initial add of game ported to PeTTa from mettalog.

It will play with only crude display of board. If you enter "m", <ENTER>, then  ( (x y) (x2 y2) ) for source and destination respectively, it will move your piece. Then a "g" command will move the AI.


patham9 (approved):
Thank you, cool example! :)

PR #72, opened 2025-11-21
Add my stdlib

rTreutlein:
Should i move all the test into an example?


rTreutlein:
Alright done. Have a look.

PR #73, opened 2025-11-24
Import via library

rTreutlein:


patham9 (changes requested):
Hm maybe that should better go into lib_prolog together with the static-import! functionality that is there to load large S expression files without functions and bangs fast. Also absolute paths are probably not needed, that isn't portable to other's computers.

patham9:
TODO library as a separate PR with test, as that feature is useful.

And we definitely can have git-import in lib-prolog once it is working (currently not).

Issue #70, opened 2025-11-19
Parentheses in quotations don't work


zariuq:
I believe the filereader.pl just looks for the next paretheses w/o checking whether there is a string or not.  Being able to quote parentheses is probably useful.

Claude Code tried to [fix it](https://github.com/zariuq/PeTTa/commit/63419fab25db42e12cc53b819d9c77ebb2d7a072), which seems to work (but probably isn't elegant enough for posterity).

Code:
```
!(println! "Testing parenthesis in string: (")
!(println! "If you see this, it worked!")
```

HE:
```
metta test_paren_he_vs_petta.metta
Testing parenthesis in string: (
If you see this, it worked!
[()]
[()]
```

PeTTa:
```
./run.sh test_paren_he_vs_petta.metta --silent
MORK init: done
Parse error: missing ')', starting at line 4:
println! "Testing parenthesis in string: (")
```

PR #74, opened 2025-11-25
Quotefix: allow semi-colons in quotes (and parentheses)

zariuq:
Cluade Code and Gemini generated.  Should at least be food for thought :).

### DIFF
--- a/examples/test_string_comments.metta
+++ b/examples/test_string_comments.metta
@@ -1,0 +1,27 @@
+; Test that semicolons inside strings are NOT treated as comments
+; This was broken before the string-aware comment stripper fix
+
+; Test 1: Semicolon constant (from set.mm)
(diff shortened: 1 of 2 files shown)


zariuq:
> That fix is quite good already, thank you very much! Will only need some investigation for compressibility as otherwise it would add 100 lines.

I'm glad that it's enough to at least make the compressed implemenation easier :)

patham9:
fixed now in https://github.com/patham9/PeTTa/commit/0c317ceb1ad75cd574f83b773f7f6999f877b9ea 
and tests by Zar: https://github.com/patham9/PeTTa/commit/686c9b5e8166dd9a5a1e7a686f5085636f032894
solves https://github.com/patham9/PeTTa/issues/70

PR #77, opened 2025-11-26
Fix Lambda FreeVar check

rTreutlein:


patham9 (approved):
Thank you, good find!

PR #78, opened 2025-11-26
Enable Direct Python File Imports

Ephrame-A:
## Overview
This PR is intended to enhance the `import!` predicate in `src/metta.pl` to support direct importing of Python (`.py`) files. This allows PeTTa to load and utilize Python modules without needing separate setup steps or complex path management.

## Changes
- **`src/metta.pl`**: 
  - Updated `import!` to detect `.py` extensions.
  - Implemented logic to resolve the absolute path of the Python file, append its directory to `sys.path`, and import the module using 
- **`examples/python_import_example.metta`**: 
  - Added a new example file demonstrating how to import a Python script and call its functions from PeTTa.

## Usage
You can now import Python scripts directly using their relative or absolute path:

```metta
; Import a python script named 'utils.py' located in the same directory (or relative to working dir)
(import! &self "utils.py")

; Or use an absolute path
(import! &self "C:/path/to/utils.py")

; Call a function from the imported module
(py-call (utils.my_function arg1))
```

#### Updated `import!` Predicate in `src/metta.pl`

```prolog
'import!'(Space, File, true) :- atom_string(File, SFile),
                                working_dir(Base),
                                ( file_name_extension(ModPath, 'py', SFile) ->
                                    absolute_file_name(SFile, Path, [relative_to(Base)]),
                                    file_directory_name(Path, Dir),
                                    file_base_name(ModPath, ModuleName),
                                    py_call(sys:path:append(Dir), _),
                                    py_call(builtins:'__import__'(ModuleName), _)
                                  ; atomic_list_concat([Base, '/', SFile, '.metta'], Path),
                                    load_metta_file(Path, _, '&self')
                                ).
```

1. Convert `File` to string `SFile` and get working directory `Base`.
2. If `SFile` has `.py` extension, resolve absolute path `Path`, extract directory `Dir` and module name `ModuleName`, add `Dir` to `sys.path`, and import the module.
3. Otherwise, append `.metta` to `SFile` and load as MeTTa file.

### DIFF
--- a/examples/python_import.metta
+++ b/examples/python_import.metta
@@ -1,4 +1,4 @@
 !(import! &self "python_import_file.py")
 
-!(test (py-call (python_import_file.greet "PeTTa User")) "Hello, PeTTa User from Python!")
+!(test (py-call (python_import_file.greet "PeTTa User")) 'Hello, PeTTa User from Python!')
 !(test (py-call (python_import_file.add 10 20)) 30)


patham9 (changes requested):
Two issues popped up:

- `load_metta_file(Path, _, '&self')` should not be there. In main branch it is `load_metta_file(Path, _, Space)`, else it can only import to `&self` after merging this PR.
- The two ! cases in python_import.metta should be turned into `!(test (py-call ...) expectedOutput)` each, else the test system sees the metta file as failing as it did not generate a test success output.

After these two are resolved I will be happy to merge it in!

Issue #82, opened 2025-12-04
!(== 1 1.0) return False


YeabsiraNigusse:
i have tested the follwoing code and it returns false
```
!(== 1 1.0)

```

i run similar code using metta and it returns True. i am not sure if it is feature or bug?

PR #83, opened 2025-12-04
Fix/equality test

YeabsiraNigusse:
# Fix for issue #82 : Numeric Equality Operator Issue

## Problem

The equality operator `==` was returning `false` when comparing integers with floats that have the same numeric value. For example:
- `== 1 1.0` returned `false` (should return `true`)
- `== 2 2.0` returned `false` (should return `true`)

## Root Cause

The equality operator in `src/metta.pl` was using Prolog's structural equality operator (`==`), which checks if two terms are structurally identical. In Prolog:
- `1 == 1.0` returns `false` because an integer and a float are different term structures
- `1 =:= 1.0` returns `true` because `=:=` checks for arithmetic equality

## Solution

Modified the equality (`==`) and inequality (`!=`) operators in `src/metta.pl` to:
1. Check if both operands are numbers
2. If both are numbers, use arithmetic equality (`=:=`) or inequality (`=\=`)
3. Otherwise, fall back to structural equality for backward compatibility with non-numeric types

## Changes Made

**File: `src/metta.pl`**

Changed from:
```prolog
'=='(A,B,R) :- (A==B -> R=true ; R=false).
'!='(A,B,R) :- (A==B -> R=false ; R=true).
```

To:
```prolog
'=='(A,B,true)  :- number(A), number(B), A =:= B, !.
'=='(A,B,false) :- number(A), number(B), A =\= B, !.
'=='(A,B,R)     :- (A==B -> R=true ; R=false).
'!='(A,B,true)  :- number(A), number(B), A =\= B, !.
'!='(A,B,false) :- number(A), number(B), A =:= B, !.
'!='(A,B,R)     :- (A==B -> R=false ; R=true).
```

## Testing

Created test files to verify the fix:
- `examples/numeric_equality_test.metta` - Tests numeric equality cases
- `examples/equality_edge_cases_test.metta` - Tests edge cases and non-numeric equality

All tests pass, including:
- ✅ `== 1 1.0` returns `true`
- ✅ `== 2 2.0` returns `true`
- ✅ `== 1 2` returns `false`
- ✅ Non-numeric equality still works correctly (atoms, strings, lists)

### DIFF
--- a/examples/equality_edge_cases_test.metta
+++ b/examples/equality_edge_cases_test.metta
@@ -1,0 +1,14 @@
+; Test that non-numeric equality still works correctly
+!(test (== "hello" "hello") True)
+!(test (== "hello" "world") False)
+!(test (== a a) True)
+!(test (== a b) False)
+!(test (== (1 2 3) (1 2 3)) True)
+!(test (== (1 2) (1 2 3)) False)
+
+; Test numeric edge cases
+!(test (== 0 0.0) True)
+!(test (== -1 -1.0) True)
+!(test (== 1.0 1) True)
+!(test (== 3.14159 3.14159) True)
+
--- a/examples/numeric_equality_test.metta
+++ b/examples/numeric_equality_test.metta
@@ -1,0 +1,14 @@
+; Test numeric equality: 1 should equal 1.0
+!(test (== 1 1.0) True)
+
+; Test that regular equality still works for same types
+!(test (== 1 1) True)
+!(test (== 1.0 1.0) True)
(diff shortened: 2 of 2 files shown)


patham9:
Changing that would require recursive checking for equality as in https://github.com/patham9/PeTTa/pull/40 which is costly. Your PR does not do that, if numbers are embedded in compounds the issue would still be there.
The following PR attempted the full recursive check: https://github.com/patham9/PeTTa/pull/40
(which hower is complex and costly, and hence also got closed)

PR #86, opened 2025-12-12
We also need to match on let in foldall

rTreutlein:


patham9 (approved):
Looks good, thank you!

PR #87, opened 2025-12-12
fix clear-space move progn/prog1 to lib_roman

rTreutlein:


patham9:
While I support getting functionality out of Translator, prog1/progn with an extra parenthesis seems very inconvenient.

Issue #109, opened 2026-01-19
HE built-in `parse` is not supported


ngeiswei:
## What is your problem?

The HE built-in `parse` that allows to take a string representing a MeTTa term and instantiating it into a MeTTa term, is not supported.

## How to reproduce your problem?

Run the following program with PeTTa:

```scheme
!(parse "(R A B)")
```

## What would you normally expect?

```
(R A B)
```

## What do you get instead?

```
(parse (R A B))
```

PR #110, opened 2026-01-19
Add parse built-in

ngeiswei:
Fix for #109.


ngeiswei:
Needless to say you should carefully check because I am not an experienced Prolog programmer.

Also, I don't know if the last test

```scheme
!(test (parse "(* 2 21)") (quote (* 2 21)))
```

is making too much commitment about both `parse` and `quote`.  Let me know if you wish me to remove it, or to add other tests, for that matter.

PR #117, opened 2026-02-01
lib_snapshot: First good story of interruptable/resumable compute

patham9:
.


patham9:
In the meanwhile already there: https://github.com/patham9/petta_lib_snapshot
and runnable without this PR merged via `git-import!`: https://github.com/patham9/petta_lib_snapshot/blob/main/snapshot_example_standalone.metta

patham9:
Here https://github.com/patham9/petta_lib_snapshot

PR #126, opened 2026-02-11
Deterministic translate_case(2)

patham9:
(commit we checked together wasn't pushed before)


patham9:
https://github.com/trueagi-io/PeTTa/pull/125

PR #141, opened 2026-03-04
Allow relative imports / functions in add/remove-atom

rTreutlein:


patham9 (changes requested):
I do not yet grasp this PR. What does it add exactly? Needs a test I guess to show? :)

patham9 (changes requested):
Translating expression on every add-atom / remove-atom call? Why is this necessary? This has significant performance impact, no?

rTreutlein:
Translating the expression for the kb is not expensive in most cases it's just a symbol but now we support functions as well.
Also it seems the notifications are getting lost. So sorry for taking so long to answer.

patham9:
Hm translating it again had significant overhead when I tried that on millions of items last time, maybe better to avoid it if possible? What does it have to do with relative imports anyways?

rTreutlein:
@patham9 What do you mean translating it again before this change whatever was in the argument position for the space was just passe as the raw symbold so you could not do:
```
(= (kb) myspace)
!(add-atom (kb) (Atom To Add)
```

And it doesn't really have anything to do with the imports to me it was just such a small clear improvment that i didn't bother with another PR.

patham9 (approved):
The addition looks legit indeed, thank you! 
I understand it now, and if nothing breaks we can have it.

PR #144, opened 2026-03-12
`Expression -> Atom` as the signal to not evaluate

meditans:
This draft is meant as a starting point for a question on the topic we talked about yesterday: this function, which should be meant as identity over syntax
```
(: foo (-> Expression Expression))
(= (foo $a) $a)
!(foo a)
```
fails because a symbol is passed.

I wanted to see what broke if I tried to change it the way MeTTa does it, and there's not much to fix: a modification to `translate_args_by_type` function suffices, the other changes are mechanical fixes the test suite.

So now for the question: what else would be impacted by this? Am I missing something? You mentioned another solution:
```
Maybe expression just shouldn't generate a typecheck for testing for expression, just leaving it unevaluated
```
What would be the upshot in this case?


patham9:
Maybe this change should happen, for the next major release v1.1.

patham9:
lib_he has conflicts @meditans  can you please resolve them? Then we can merge thi PR.

patham9:
Thank you! Will finish this PR in https://github.com/trueagi-io/PeTTa/pull/212

PR #148, opened 2026-03-24
Add list deconstruction similar to Lisp for function arguments.

Habush:
- Adds list deconstruction similar to Lisp with `.` operator. This will provide a syntactic sugar of the form `(= (sum ($head . $rest) )...))`
- Add `is-ground/1` predicate for checking if an expression is a ground expression. A complement to the already existing `is-var/1`  predicate.

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -159,6 +159,7 @@ get_type_candidate(X, T) :- match('&self', [':',X,T], T, _).
 'get-metatype'(X, 'Symbol') :- atom(X), !.            % e.g., a
 
 'is-var'(A,R) :- var(A) -> R=true ; R=false.
+'is-ground'(A,R) :- ground(A) -> R=true ; R=false.
 'is-expr'(A,R) :- is_list(A) -> R=true ; R=false.
 'is-space'(A,R) :- atom(A), atom_concat('&', _, A) -> R=true ; R=false.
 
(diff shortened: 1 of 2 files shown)


patham9 (changes requested):
Hm, sure it should be infix?
In what way is it better than the already supported `(cons $head $tail)` pattern?
A testcase would be needed to see I guess.

patham9:
Thank you though! Can we have is-ground by itself for now in this PR @Habush ? Then I am happy to merge it

Habush:
> In what way is it better than the already supported (cons $head $tail) pattern?

It was a style choice as it is common list deconstruction pattern in Lisp-like languages. But I understand if infix operators don't match PeTTa's style. I have removed the operator and the PR now contains only `is-ground`.

patham9 (changes requested):
Hm is-ground also needs to be registered I suppose? Do you have a testcase for it?

patham9:
@Habush if you register the function we can merge the PR

Habush:
Registered the function. Please check

patham9:
It also removes is-alpha-member, why?

Habush:
Oops, that was by mistake. I'll close this PR and create a now for a cleaner commit history

Habush:
Closed in favor #172

PR #151, opened 2026-03-27
# Add alpha-unique-atom: Alpha-Equivalence Deduplication Function

### LINKED ISSUE #149: unique-atom does not remove alpha-equivalent atoms
# Bug: unique-atom does not handle alpha-equivalence correctly

## Description
The current implementation of `unique-atom` in PeTTa only removes **syntactically identical atoms**, but does not consider **alpha-equivalence** (structural equality with variable renaming).  

As a result, atoms that are structurally identical but differ only by variable names are **not treated as duplicates**, leading to unexpected duplicate entries.  

Currently, there is **no built-in function in PeTTa to handle this alpha-equivalence deduplication**, which is often needed in symbolic reasoning tasks.

## Steps to Reproduce
Run `unique-atom` on a list of atoms that are structurally identical but use different variable names. For example:

```lisp
(unique-atom ((link $x human) (link $y human) (link $z human)))

Observe that all atoms are kept, even though they are alpha-equivalent.

**Expected Behavior**

unique-atom should treat alpha-equivalent atoms as duplicates and keep only one representative.

**Actual Behavior**

All alpha-equivalent atoms are kept as if they are unique, e.g.:

((link $x human) (link $y human) (link $z human))


**Additional Context**

This issue affects deduplication in symbolic reasoning and pattern generation tasks. Without proper handling of alpha-equivalence, downstream logic can produce incorrect or redundant results.

There is currently no built-in solution in PeTTa for this problem, which motivates the need for an additional function like alpha-unique-atom to safely remove alpha-equivalent duplicates with out affect the exiting unique-atom

### PR DESCRIPTION
**Pull request by aprilyab**
## Summary

This PR introduces a new built-in function, `alpha-unique-atom`, to the PeTTa language. Unlike the existing `unique-atom`, which only removes syntactically identical atoms, `alpha-unique-atom` deduplicates atoms using alpha-equivalence (structural equality with variable renaming). This ensures that atoms differing only by variable names are treated as duplicates and only one representative is kept.

## Problem and reproduction

The current `unique-atom` function in PeTTa does not consider alpha-equivalence, leading to unexpected duplicates when atoms differ only by variable names. This causes issues in symbolic reasoning and pattern mining tasks, where structurally identical atoms should be considered equal.

**Example:**

Input:
```metta
!(alpha-unique-atom ((link $x human) (link $y human) (link $z human)))
```
Observed (incorrect) output with unique-atom:
```metta
((link $x human) (link $y human) (link $z human))
```
Expected output with alpha-unique-atom:
```metta
((link $a human))
```

## Root cause

`unique-atom` only checks for syntactic equality, not structural (alpha) equivalence. As a result, atoms that are structurally identical but use different variable names are not deduplicated.

## What I changed

- Added `alpha-unique-atom` as a new built-in, implemented in Prolog and registered in the PeTTa core.
- The function uses structural equality (`=@=`) to compare atoms, ensuring alpha-equivalent atoms are deduplicated.
- All other behavior remains unchanged; only the equality notion for deduplication is altered.

## Why this fix

Alpha-equivalence is the correct equality notion for deduplication in symbolic reasoning: two atoms that only differ by the names of their variables are treated as the same pattern, while distinct syntactic patterns are preserved.

## Testing & validation

A comprehensive test suite is included, covering:
- Basic duplicates with different variables
- Different functors and nested structures
- Mixed unique and duplicate elements
- Numbers, atoms, empty and single-element lists

Example test:
```metta
!(test (=alpha (alpha-unique-atom ((link $x human) (link $y human) (link $z human)))
               ((link $a human)))
       True)
```

## Impact

- Enables correct deduplication for symbolic reasoning and pattern mining.
- Does not change the behavior of the existing `unique-atom` function, preserving backward compatibility for projects relying on its current semantics.

## Related issues

Fixes [unique-atom does not remove alpha-equivalent atoms#149](https://github.com/trueagi-io/PeTTa/issues/149)

### DIFF
--- a/examples/test_alpha_unique_atom.metta
+++ b/examples/test_alpha_unique_atom.metta
@@ -1,0 +1,59 @@
+; tests for alpha-unique-atom 
+
+; 1 Basic duplicates with different variables
+!(test (=alpha (alpha-unique-atom ((link $x human) (link $y human) (link $z human)))
+               ((link $a human)))
+       True)
+
+!(test (=alpha (alpha-unique-atom ((parent $x human) (parent $y human) (child $z human)))
+               ((parent $a human) (child $b human)))
+       True)
+
+; 2 Different functors
+!(test (=alpha (alpha-unique-atom ((parent $x human) (child $y human) (friend $z human)))
+               ((parent $a human) (child $b human) (friend $c human)))
+       True)
+
+!(test (=alpha (alpha-unique-atom ((likes $x) (hates $y) (knows $z)))
+               ((likes $a) (hates $b) (knows $c)))
+       True)
+
+; 3 Nested structures
+!(test (=alpha (alpha-unique-atom ((link (foo $x) human) (link (foo $y) human) (link (bar $z) human)))
+               ((link (foo $a) human) (link (bar $b) human)))
+       True)
+
+!(test (=alpha (alpha-unique-atom ((parent (child $x) human) (parent (child $y) human) (parent (child $x) human)))
+               ((parent (child $a) human)))
+       True)
+
+; 4 Mix of unique and duplicates
+!(test (=alpha (alpha-unique-atom ((link $x human) (parent $x human) (link $y human) (parent $z human) (link $x human)))
+               ((link $a human) (parent $a human)))
+       True)
+
+!(test (=alpha (alpha-unique-atom ((foo $x) (foo $y) (bar $x) (foo $x) (bar $y)))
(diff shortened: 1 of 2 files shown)


patham9 (approved):
Well-done and well-tested! Thank you so much!!

rTreutlein:
While i agree that this is an important function. I would generally prefer such things to be implemented in MeTTa directly.

PR #155, opened 2026-04-02
Refactor CLI and filereader to allow PeTTa to emit Prolog code as intermediate step.

rmgaray:
PeTTa currently cannot be used as a transpiler because it does not output the generated Prolog clauses when translating MeTTa code. It also always executes all goals, which is not what we want when running it as a compiler. This PR fixes that:

- [x] Modified `load_metta_file` and other clauses to return the generated Prolog program in addition or instead of the execution results.
- [x] Added new runtime option to disable MeTTa code execution and refactored `process_form` to use it.
- [x] Modified the `filereader` module to take into account the already existing runtime option `silent` and disable the debug messages when interpreting/compiling.
- [x] Removed ad-hoc option parsing in `fileloader.pl` and introduced a more principled CLI using the `optparse` library. PeTTa has now two modes of operation: compiler and interpreter.
- [x] Generated 'add-atom' clauses for all MeTTa sexprs when compiling MeTTa programs.
- [x] Add runtime clauses to generated output
- [x] Modified `translate_expr` to print the results of each runnable. This is achieved by optionally adding a write clause at the end of each runnable body. This is only done for top-level MeTTa forms and whenever we are in compiler mode, not when interpreting MeTTa programs the usual way.
- [ ] Make compiled clauses dynamic (we need this to fully support MeTTa).
- [ ] Run all tests in compiled mode?

### DIFF
--- a/examples/builin_types.metta
+++ b/examples/builin_types.metta
@@ -1,4 +1,4 @@
-!(import! &self ../lib/lib_builtin_types)
+!(import! &self lib/lib_builtin_types)
 
 ;; Test type definitions of arithmetic operators
 !(test (get-type +) (-> Number Number Number))
--- a/examples/curry.metta
+++ b/examples/curry.metta
@@ -1,16 +1,3 @@
 (= (f $a $b) (+ $a $b))
 
-(= (g $a $b $c) (+ $c (+ $a $b)))
-
-(= (show) (repr (f 1)))
-
 !(test (repr (f 1)) "(partial f (1))")
-!(test ((f 1) 2) 3)
-!(test (repr (g 1 2)) "(partial g (1 2))")
-
-(= (h $A $B)
-   (append ($A) $B))
-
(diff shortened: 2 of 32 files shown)


patham9:
Thank you for your effort! Closed due to inactivity for now. However having a branch that supports this might still be useful for some application even when not up-to-date with main, was this already working to some degree? If so I will reference it from the Wiki at least.

patham9:
Renamed as the title was misleading.

rmgaray:
> Thank you for your effort! Closed due to inactivity for now. However having a branch that supports this might still be useful for some application even when not up-to-date with main, was this already working to some degree? If so I will reference it from the Wiki at least.

I believe most tests of the PeTTa suite were green in this branch, but roughly 10-15% of the tests were not passing. These last tests were very tricky, often involving function specialization and/or type checking. I have the feeling that there might be some fundamental flaw in this approach (because of how often fixing these tests ended up breaking others), but I cannot say for sure, to be honest.

PR #157, opened 2026-04-03
Fix/duplicate import error throwing

ermijeremy:
## Related Issue

Closes #154

## Summary
Prevent duplicate `import!` loads by tracking absolute import paths and ignoring duplicate imports.

## What Changed
- `import!` now normalizes every import to an absolute path.
  - `.py` imports resolve to an absolute path and import the Python module.
  - Other imports resolve to an absolute `.metta` path and load the MeTTa file.
- Each imported absolute path is stored in a dynamic predicate `imported_file/1`.
- If the same absolute path is imported again, it will be ignored`.

## Why This Fixes the Issue
Previously, repeated `import!` calls could reload the same file multiple times, causing duplicated definitions and inconsistent behavior.  
By normalizing to a canonical absolute filename and recording it in `imported_file/1`, any subsequent import of the same file is detected and rejected reliably.

## How It Works (Mechanics)
1. Resolve `File` to an absolute path.
2. Check `imported_file(AbsPath)`:
   - If found → ignore the duplicated import.
   - If not → load file and `assertz(imported_file(AbsPath))`.

## Tests
- Added minimal test that imports the same file twice in one MeTTa file and asserts an exception is raised.

### DIFF
--- a/examples/add_atom_fun_space.metta
+++ b/examples/add_atom_fun_space.metta
@@ -1,0 +1,6 @@
+
+
+(= (space) my_space_name)
+
+!(add-atom (space) (my test atom))
+!(test (match (space) $a $a) (my test atom))
--- a/python/helper.pl
+++ b/python/helper.pl
@@ -2,6 +2,7 @@ maybe_enable_silent(true) :- !.
 maybe_enable_silent(false) :- assertz(silent(true)).
 
 set_working_dir(load_metta_file, File) :-
+    clear_imported_files,
     file_directory_name(File, Dir),
     retractall(working_dir(_)),
     assertz(working_dir(Dir)),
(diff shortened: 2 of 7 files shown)


patham9 (changes requested):
Thank you very much!
Only: there should be no import error being generated.
Often libraries depend on same core libraries, it is normal.
It only should ensure that no library is loaded twice.

ermijeremy:
> Thank you very much! Only: there should be no import error being generated. Often libraries depend on same core libraries, it is normal. It only should ensure that no library is loaded twice.

@patham9 it no longer throws error instead if a duplicate import is found it just ignores it 
-> true

ermijeremy:
> Thank you very much! Only: there should be no import error being generated. Often libraries depend on same core libraries, it is normal. It only should ensure that no library is loaded twice.

@patham9 @pisaev1 any updates!!

patham9:
Thank you for your patience and I will take a look asap. There are multiple PR's that affect import handling, will try to get an overview first before merging.

patham9:
Superseded by #203

PR #159, opened 2026-04-04
Fix import! to throw error when target file/library is not found

ermijeremy:
### Summary
This PR changes the import execution path so missing files/library targets raise a real Prolog exception instead of failing by backtracking to `false`.

### Root Cause
`import!` previously swallowed all importer failures:

```prolog
'import!'(Space, File, true) :- catch(importer_helper(Space, File), _, fail).
```

### Implementation Details
The change removes blanket exception swallowing and makes failure explicit.

`import!` now directly delegates to `importer_helper/2`:

```prolog
'import!'(Space, File, true) :- importer_helper(Space, File).
```

resolution now follows:

1. Build candidate path (`Path`)
2. Normalize extension with `ensure_metta_ext/2` -> `PathWithExt`
3. If file exists, load via `load_metta_file/3`
4. Otherwise, throw:

```prolog
throw(error(existence_error(source_sink, File),
            context('import!', File)))
```

This yields standard SWI-Prolog error semantics and is catch-compatible.

For valid imports, behavior is unchanged:
     path resolution
     `.metta` extension normalization
     loading into the target space

#### Before
     Missing import: silent fail (`false`)

#### After
    Missing import: explicit runtime error (`existence_error(source_sink, ...)`)
    `catch` receives a concrete error term and can handle/report it

### Validation
    `!(catch (import! ...missing...))` returns structured `Error(...)` value as expected.

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -1,6 +1,6 @@
 %%%%%%%%%% Dependencies %%%%%%%%%%
-library(X, Path) :- library_path(Base), atomic_list_concat([Base, '/', X], Path).
-library(X, Y, Path) :- library_path(Base), atomic_list_concat([Base, '/../', X, '/', Y], Path).
+library(X, Path) :- library_path(Base), atomic_list_concat([Base, '/', X], Path), !.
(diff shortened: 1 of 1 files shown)


patham9:
Superseded by #203.

PR #161, opened 2026-04-07
Make Importing Order Independence

aprilyab:
## Summary

This PR implements **order-independent importing** in PeTTa. The import system and runtime have been enhanced so that all function definitions and dependencies are registered before evaluation, allowing files to be imported in any order without affecting program correctness. This eliminates a major source of user error.

## Problem and reproduction

Previously, importing files in the wrong order could cause missing function errors or unevaluated calls, because function signatures were only registered as files were loaded. This made modular code fragile and error-prone.

### Example Files

**utils.metta**
```metta
(= (is-even $x) (== (% $x 2) 0))
```

**dependent.metta**
```metta
(= (process-number $x)
   (if (is-even $x)
	   (+ $x 20)
	   (* $x 3)))
```

**main.metta**
```metta
!(import! &self dependent)   ; Import the dependent module first (which uses is-even)
!(import! &self utils)       ; Import the utility module second (which defines is-even)

!(process-number 30)
```

### Result Before the Fix (Order-Dependent)
- `main` imports `dependent` before `utils`, so `is-even` is not defined when `process-number` is compiled.
- The fallback is always the "else" branch: `(* 30 3) = 90`
- **Output:**
	```
	90
	```

### Result After the Fix (Order-Independent)
- All function signatures are registered before evaluation, so `is-even` is available to `process-number` regardless of import order.
- `is-even 30` is true, so the "then" branch: `(+ 30 20) = 50`
- **Output:**
	```
	50
	```

## Root cause

The import system registered function signatures only as files were loaded, so dependent functions were not visible if their defining file was imported later. This led to order-dependent errors and brittle code.

## What I changed

- Enhanced the import system to pre-register all function signatures from all imported files before any code is compiled or evaluated.
- Updated the runtime to ensure all dependencies are available regardless of import order.

## Why this fix

Order independence is essential for modularity, maintainability, and user experience. This change allows users to organize and import their code in any order, without worrying about dependency errors or unevaluated calls. It brings PeTTa in line with modern language best practices.

## Testing & validation

A comprehensive test suite is included to validate the new behavior:
- Imports dependent and helper modules in any order
- Verifies correct evaluation of functions that depend on cross-file definitions

## Impact

- Guarantees that import order does not affect program correctness
- Enables robust, modular code organization in PeTTa


patham9:
What if imports are done in an imported file? How can order-independence still be guaranteed? And in such case, will it still work if there are cyclic imports? Related to https://github.com/trueagi-io/PeTTa/pull/157

aprilyab:
> What if imports are done in an imported file? How can order-independence still be guaranteed? And in such case, will it still work if there are cyclic imports? Related to #157

What if imports are done in an imported file? How can order-independence still be guaranteed?
Order-independence is guaranteed even for nested imports (imports are done in an imported file) because the system recursively pre-registers all function signatures from every imported file—directly or indirectly—before any code is evaluated. This ensures that all dependencies are available, regardless of the import order or depth.

And in such case, will it still work if there are cyclic imports?
Yes, with the addition of new logic, cyclic imports are also now supported. The system tracks which files have already been processed and skips re-importing them, preventing infinite recursion. This means that even if two or more files import each other (directly or indirectly), all their functions are registered and available, and the program runs without errors or hangs. Previously, cyclic imports would cause issues, but this is now fully handled.

Test cases for all these scenarios—import order independence, nested imports, and cyclic imports—are included and verified in the examples directory.

patham9:
Superseded by #203

PR #162, opened 2026-04-07
Fix/is member

Tewobsta-sa:
<h1>Add <code>is-alpha-member</code> function for proper variable membership checking</h1>
    <h2>Summary</h2>
    <p>
        This PR adds a new <code>is-alpha-member</code> function that properly handles membership checking for unassigned variables in patterns. 
        It fixes an issue where the existing <code>is-member</code> function incorrectly returns multiple <code>true</code> results when 
        checking membership of undefined variables.
    </p>
    <h2>Problem</h2>
    <p>
        The current <code>is-member</code> function treats unassigned variables as wildcards, causing it to match every element in a list.
    </p>
    <blockquote>
        <strong>Example:</strong> When checking <code>(is-member $new (hi name boss))</code>, it returns <code>true</code> three times 
        instead of the expected single <code>false</code> result.
    </blockquote>
    <h2>Solution</h2>
    <p>
        Implemented <code>is-alpha-member</code> using alpha-equivalence checks to ensure unassigned variables are not treated as wildcards. 
        The new function:
    </p>
    <ul>
        <li>Uses <code>member_alpha</code> with <strong>strict equality</strong> (<code>X = H</code>) instead of unification.</li>
        <li>Returns a <strong>single boolean result</strong> (true/false) instead of multiple matches.</li>
        <li>Properly handles variable scoping during pattern matching.</li>
    </ul>
    <h2>Implementation</h2>
<pre>
member_alpha(X, [H|_]) :- (var(X) -> var(H) ; true), X = H, !.
member_alpha(X, [_|T]) :- member_alpha(X, T).
'is-alpha-member'(X, List, true) :- member_alpha(X, List), !.
'is-alpha-member'(_, _, false).
</pre>
    <h2>Testing</h2>
    <p>The fix resolves the issue demonstrated by the following test case:</p>
<pre>
!(let*
    (
        ($pat (hi name boss))
        ($dummy1 (println! (pattern:- $pat)))
        ($bool (is-alpha-member $new $pat))
        ($dummy2 (println! (is member:- $bool in pattern:- $pat)))
    )
    ()
)
</pre>
    <p><strong>Correct Output:</strong> <code class="success">(is member:- false in pattern:- (hi name boss))</code></p>

### DIFF
--- a/examples/is_alpha_member_test.metta
+++ b/examples/is_alpha_member_test.metta
@@ -1,0 +1,54 @@
+;;; Test file for member_alpha and is-alpha-member predicates  
+;;; This file tests edge cases for alpha-equivalence membership checking  
+  
+;;; Test 1: Empty list should return false  
+!(test (is-alpha-member x ()) false)  
+  
+;;; Test 2: Variable in list should unify  
+!(test (is-alpha-member $x (a b c)) true)  
+  
+;;; Test 3: Ground term membership  
+!(test (is-alpha-member a (a b c)) true)  
+!(test (is-alpha-member d (a b c)) false)  
+  
+;;; Test 4: Alpha-equivalence with variables  
+!(test (is-alpha-member (f $x) ((f $y) (g $z)) ) true)  
+!(test (is-alpha-member (f $x) ((f $y) (f $y)) ) true)  
+  
+;;; Test 5: Complex nested structures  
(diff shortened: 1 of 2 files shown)


patham9 (approved):
Looks good!

PR #171, opened 2026-05-09
Allow relative imports / functions in add/remove-atom

patham9:
Re-opened your PR @rTreutlein 
As it was it broke OmegaClaw imports for me so I rolled back for now, will investigate.


rTreutlein:
If you have any information about what broke i can fix it as well.

patham9:
Close as there is a new PR as I understand: https://github.com/trueagi-io/PeTTa/pull/195

PR #172, opened 2026-05-12
Add is-ground function as a counter-part to is-var

Habush:
- Add is-ground/1 predicate for checking if an expression is a ground expression. A complement to the already existing is-var/1 predicate


patham9 (approved):
Awesome, thank you!

PR #173, opened 2026-05-17
Add type-casting Support

AlexKalll:
This PR implements type-casting support and wires type checking by the tests after casting. 

- Adds a `type-cast` helper function in `lib_he.metta`, enables a `get-types` invocation/call in `he_types.metta` after type is casted by preferring space-registered types via match in `metta.pl`.  simply old type declarations are overridden in the space when a new type is cast.

### DIFF
--- a/examples/he_types.metta
+++ b/examples/he_types.metta
@@ -2,10 +2,11 @@
 
 !(test (is-function (-> Atom Atom)) True)
 !(test (is-function Atom) False)
-;TODO:
-;(: type1 Type)
-;!(type-cast A type1 &self)
-;!(type-cast 1 type1 &self)
+(: type1 Type)
+!(type-cast A type1 &self)
+!(type-cast 1 type1 &self)
+!(test (get-type A) type1)
+!(test (get-type 1) type1)
 !(test (match-types Atom Atom "Matched!" "Didn't match") "Matched!")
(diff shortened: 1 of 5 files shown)


patham9:
A variable can get instantiated with any value, hence its type should not be %Undefined% but a variable as well.
Also not sure if we should change the type system that much just to extend the lib_he compatibility lib.
If typecasting is useful more generally, maybe we can consider it.
Sometimes I wonder if type require casting, why specify types at all?

Please feel free to discuss and reopen if you think it matters, I did not yet fully grasp the need for this.

AlexKalll:
> A variable can get instantiated with any value, hence its type should not be %Undefined% but a variable as well. Also not sure if we should change the type system that much just to extend the lib_he compatibility lib. If typecasting is useful more generally, maybe we can consider it. Sometimes I wonder if type require casting, why specify types at all?
> 
> Please feel free to discuss and reopen if you think it matters, I did not yet fully grasp the need for this.

Your point about variable types being variables rather than %Undefined% makes sense. I added type-casting mainly to activate the todo tests in he_types.metta, but I agree the type system changes might not be important as long as we can instantiate a variable with any value. 

I will leave this closed for now and will revisit if a clearer use case comes up. Thanks @patham9 for the thoughtful review.

PR #175, opened 2026-05-24
Fix quote-aware parsing in lib_import

Aryam-21:
This PR improves the parsing logic in `lib_import.pl` to correctly handle quote-aware imports and preserve quoted content during token parsing.

The update focuses on making import parsing more reliable while keeping the existing behavior intact for other parsing cases.

### DIFF
--- a/lib/lib_import.pl
+++ b/lib/lib_import.pl
@@ -18,12 +18,32 @@ convert_stream(In, Out, Space) :- read_line_to_string(In, Line),
 
 %Perform simple transformation from S-Expression to space-rel predicate:
 convert_line(Line0, Space, Out) :- sub_string(Line0, 1, _, 1, Inner0),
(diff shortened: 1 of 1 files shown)


patham9 (changes requested):
Thank you! Looks good but needs cuts else the last transform_chars clause matches always no?

Aryam-21:
Thanks for the feedback! I've added ( ! ) cuts to all the specific transform_chars clauses so the catch-all can no longer match on backtracking.

patham9:
Looks good. What is the performance impact for loading a large database like the bio one (first removing the generated .pl and generated compiled file)? If it does not have much impact we can merge it.

patham9 (approved):
Since its only in the preprocessing step and the loading of the compiled variant is what matters, I think we can already merge it actually.

PR #177, opened 2026-05-29
Omegaclaw rust native 4960819518872610031

autonull:


patham9:
Interesting, thank you! It contains MeTTa execution on top of pathmap? Maybe that would be best-suited as a PR to PeTTa? Closing for now. And which MORK/Pathmap commit/branch does it rely on?

PR #179, opened 2026-06-04
Add alpha-unique that operates on a superposition

ngeiswei:
It is like `alpha-unique-atom` but operates on a superposition instead of a tuple.

I did not find it necessary to add tests because `alpha-unique-atom` is already extensively covered.


patham9 (approved):
Looks good, thank you!

PR #183, opened 2026-06-11
fix: make car-atom/cdr-atom safe on empty or non-lists & support short-circuit operators

### LINKED ISSUE #182: Bug involving laziness and `car-atom`
## What is your problem?

A function combining `and` and `car-atom` called on a symbol seems to prematurely prune the branch due to `car-atom` expecting an expression.

## How to reproduce your problem?

Run the following code in PeTTa

```scheme
;; Return True iff the given term is a lambda abstraction, meaning an
;; expression of the form (λ VARIABLES BODY)
(= (is-lambda $x)
   (and (== (get-metatype $x) Expression)
        (and (not (== () $x)) (== λ (car-atom $x)))))

;; Test is-lambda
!(test
  (is-lambda A)
  False)
!(test
  (is-lambda (A B))
  False)
!(test
  (is-lambda (λ $x $y))
  True)
```

## What would you normally expect?

```
is false, should false. ✅ 
is false, should false. ✅ 
is true, should true. ✅ 
true
true
true
```

## What do you get instead?

```
is (), should false. ❌ 
```

## What else do have to say?

Replacing

```scheme
(= (is-lambda $x)
   (and (== (get-metatype $x) Expression)
        (and (not (== () $x)) (== λ (car-atom $x)))))
```

by

```scheme
(= (is-lambda $x)
   (if (== (get-metatype $x) Expression)
       (if (== () $x)
           False
           (== λ (car-atom $x)))
       False))
```

fixes the issue, indicating that it might have something to do with `and`, maybe its laziness, or such.

AlexKalll:
This PR:
Fixes [car-atom & cdr-atom on empty input cases](https://github.com/trueagi-io/PeTTa/issues/182) by making the `car-atom` and `cdr-atom`  return `()` on empty or non-list inputs instead of crashing.

Additionally, Adds `and-then` and `or-else` to preserve the short-circut approach alongside the car-atom and cdr-atom fixes.(i.e e.g if the first operand is False, no need to evalute the second one in the case os `and-then` and if one is true, no need of evaluating the second in the case of `or-else`) .

### DIFF
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -139,7 +139,8 @@ alpha_list_to_set_assoc([H|T], SeenIn, R) :-
 'size-atom'(List, Size) :- length(List, Size).
 'car-atom'([H|_], H) :- !.
 'car-atom'(_, []).
-'cdr-atom'([_|T], T).
+'cdr-atom'([_|T], T) :- !.
+'cdr-atom'(_, []).
 decons([H|T], [H|[T]]).
 cons(H, T, [H|T]).
 'index-atom'(List, Index, Elem) :- nth0(Index, List, Elem).
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -137,7 +137,8 @@ alpha_list_to_set_assoc([H|T], SeenIn, R) :-
 
 'sort-atom'(List, Sorted) :- msort(List, Sorted).
 'size-atom'(List, Size) :- length(List, Size).
-'car-atom'([H|_], H).
+'car-atom'([H|_], H) :- !.
+'car-atom'(_, []).
 'cdr-atom'([_|T], T).
 decons([H|T], [H|[T]]).
 cons(H, T, [H|T]).
--- a/src/translator.pl
+++ b/src/translator.pl
@@ -174,6 +174,13 @@ translate_expr([H0|T0], Goals, Out) :-
                                                      ; translate_expr(KeyExpr, Gk, Kv),
                                                        translate_case(PairsExpr, Kv, Out, IfGoal, KeyGoal),
                                                        append([GsH, Gk, KeyGoal, [IfGoal]], Goals) )
+        %--- Short-circuit boolean operators ---:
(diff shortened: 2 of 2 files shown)


patham9:
Boolean operators are explicitly kept to be relational
E.g.
`!(($x $y) (and $x $y))`
generates the entire truth table.
Not sure we want short-circuit by default as it takes away some logic power.

AlexKalll:
> Not sure we want short-circuit by default as it takes away some logic power.

Thank you for reviewing & the feedback, 
yeah I see that the short-circuit approach would break that.

so instead of modifying and/or, what if the `car-atom` shall return `()` on `non-lists` instead of failing. that way would keep its relational behavior. 

what do you think of this approach? @patham9

AlexKalll:
> Boolean operators are explicitly kept to be relational E.g. `!(($x $y) (and $x $y))` generates the entire truth table. Not sure we want short-circuit by default as it takes away some logic power.

I  pushed this approach. `car-atom` now return () on empty/non-lists instead of crashing.  so `and` e.g !(($x $y) (and $x $y)) generates the entire truth table. could you review & test it again?
thank you.

patham9:
This sounds reasonable! But I also like your short-cut and and short-cut or, maybe they should also exist as
```
(and-then A B)  ; A, then B only if needed
(or-else A B)   ; A, else B only if needed
```
in the translator, in addition to the new behavior for car-atom and cdr-atom.

AlexKalll:
> This sounds reasonable! But I also like your short-cut and and short-cut or, maybe they should also exist as

> (and-then A B)  ; A, then B only if needed
> (or-else A B)   ; A, else B only if needed

> in the translator, in addition to the new behavior for car-atom and cdr-atom.

Great, I will add and-then and or-else to preserve the short-circut approach alongside the car-atom and cdr-atom fixes. Thank you

patham9 (approved):
Thank you, looks great!

PR #184, opened 2026-06-17
fix on get-state function

YeabsiraNigusse:
## Summary

Fix `get-state` so requesting an undefined state initializes it with an empty value instead of raising an error.

## Problem

Previously, calling `get-state` on a state that had not been created with `new-state` caused an error because the implementation directly used `nb_getval/2`. This made expressions like:

```metta
!(test (get-state fresh-state) ())
```
fail when fresh-state did not already exist.

## Solution
- Updated get-state to:
- Check whether the state name is an atom.
- Return the existing stored value when the state already exists.
- Create the state with an empty value () when it does not exist yet.
- Preserve the previous behavior for non-atom cases.
- This makes get-state behave more safely and intuitively for newly referenced states.

### DIFF
--- a/examples/state.metta
+++ b/examples/state.metta
@@ -2,3 +2,4 @@
 !(test (get-state state) rest)
 !(change-state! state active)
 !(test (get-state state) active)
+!(test (get-state fresh-state) ())
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -239,7 +239,16 @@ assert(Goal, true) :- ( call(Goal) -> true
 %%% States: %%%
 'bind!'(A, ['new-state', B], C) :- 'change-state!'(A, B, C).
 'change-state!'(Var, Value, true) :- nb_setval(Var, Value).
-'get-state'(Var, Value) :- nb_getval(Var, Value).
+'get-state'(Var, Value) :-
+    ( atom(Var) ->
+        ( nb_current(Var, Stored) ->
+            true
+        ;   Stored = [],
+            nb_setval(Var, Stored)
+        ),
+        Value = Stored
+    ;   nb_getval(Var, Value)
+    ).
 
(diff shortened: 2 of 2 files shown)


patham9:
Thank you! Makes sense but I am not yet sure about the performance impact of this, I guess it isn't affecting it much as it should be like a hashmap lookup / O(1) to check the existence under the cover? But might be a constant of 2x hidden there, maybe good to check.

YeabsiraNigusse:
> Thank you! Makes sense but I am not yet sure about the performance impact of this, I guess it isn't affecting it much as it should be like a hashmap lookup / O(1) to check the existence under the cover? But might be a constant of 2x hidden there, maybe good to check.

I checked this with a small SWI-Prolog microbenchmark over 50M calls.

For an already-existing state:
- old `get-state` using `nb_getval/2`: `6.946s`, ~`138.93 ns/call`
- new `get-state` using `atom/1` + `nb_current/2` + default initialization fallback: `8.707s`, ~`174.14 ns/call`

So the new path is about `1.25x` in this tight benchmark, with an absolute overhead of ~`35 ns/call`. The raw primitives were close (`nb_getval/2` ~`131 ns/call`, `nb_current/2` ~`138 ns/call`), so most of the extra cost is the guarded existence/default handling rather than a large lookup penalty.

The behavior difference is that the old version throws on a missing state, while the new version initializes missing atom states to `[]` and succeeds.

[Here](https://github.com/YeabsiraNigusse/PeTTa/blob/benchmark/benchmarks/bench_get_state_minimal.pl) is the link for the benchmark code

patham9:
Not sure if the corner case of forgetting to set a state justifies general 25% performance penalty here.
H-E returns an error as well in this case.
Closed for now but happy to reopen and discuss further if you want.

PR #186, opened 2026-06-28
fix: Atom arg type and case Empty default works(#185)

### LINKED ISSUE #185: Atom argument type doesn't prevent argument from being evaluated
The following example is expected to return both `Ok` and `Nok` but it returns only `Nok`:
```metta
(: foo (-> Atom %Undefined%))
(= (foo $x)
   (case $x
     ((Empty Ok)
      ($_ Nok))))
!(foo (empty))
!(foo 1)
```

AlexKalll:
This PR fixes #185. Two related fixes: (1) The `Atom` arg type was evaluating the argument instead of passing it as data, unlike `Expression`. So `(foo (empty))` generated `empty(_)` which always fails, and the `foo` function was never called. Now `Atom` is treated like `Expression`. that means, the argument is passed unevaluated. (2) The `Empty` default in `case` only fired/run when the key expression itself failed to evaluate, making it dead/unreachable code for variable keys(because the wildcard `$_` always matches with anything). After the fix, now `Empty` also matches when the key value is the empty value `(empty)` by introducing `== [empty] ` checking mechanism in the translator.

additionally, all the existing test passes after the fix, including the original bug below, which is identified in the [issue](https://github.com/trueagi-io/PeTTa/issues/185)
```python
(: foo (-> Atom %Undefined%))
(= (foo $x)
   (case $x
     ((Empty Ok)
      ($_ Nok))))

!(test (foo (empty)) Ok)
!(test (foo (1)) Nok)
```


vsbogd (commented):
Thanks @AlexKalll for this change! It works indeed and now I know that I can use `Expression` type as a workaround. Two comments from me on this:

1. I think it makes sense to add the test case into the list of tests
2. It is a bit strange that if we exchange case branches the test still passes:
```metta
(= (foo $x)
   (case $x
     (($_ Nok)
      (Empty Ok))))

!(test (foo (empty)) Ok)
```
It would be more logical to have default branch (if it is first) be matched first. But I am not sure if it can be easy enough implemented so I am considering this minor issue. Anyway it is an improvement. Thanks!

It is also interesting what @patham9 thinks about this.

AlexKalll:
> Thanks @AlexKalll for this change! It works indeed and now I know that I can use `Expression` type as a workaround. Two comments from me on this:


Thanks @vsbogd for the careful review! Yes, that is right, my fix makes `Empty` always match first, regardless of where it appears in the case. 

One thing I noticed:  the fix only works with  type declaration (Atom or Expression). Without it, the argument gets evaluated and the function is never called. So I think the type declaration is needed to trigger the "pass as data" behavior.

I also added the test case to examples/ (case_empty.metta) module including the case-order behavior and all passed. Looking forward to @patham9's thoughts as well. Thank you!

vsbogd:
> So I think the type declaration is needed to trigger the "pass as data" behavior.

Yes, this is expected behavior.

AlexKalll:
> Yes, this is expected behavior.

Got it, Thank you for confirming!

patham9:
I disagree. (empty) cannot be passed to a function.
The compiled output creates an explicit [empty] check which is invalid.

The previous behavior was correct.

PR #192, opened 2026-07-08
fix: make non-editable pip install of the petta package work (#190)

### LINKED ISSUE #190: pip install . (non-editable) installs a petta package missing the petta module
setup.py uses packages=find_packages(where="python") with package_dir={"": "python"}. python/petta.py is a loose module file, not a package (there's no python/petta/__init__.py), and find_packages() only discovers directories that contain __init__.py. It can't see a bare .py module sitting next to them. The only thing under python/ it does pick up is python/tests/, which has its own __init__.py.

A plain pip install . from a clean clone builds and installs a petta distribution whose wheel contains tests/* and the dist-info, but not petta.py itself. import petta then fails with ModuleNotFoundError in an environment that only has the installed package on sys.path.

pip install -e . works around this, since an editable install adds the source tree to sys.path directly instead of relying on packages/py_modules to decide what to copy. That's likely why this hasn't been caught: a dev checkout that installs editable never hits it.

I ran into this installing this repo as a dependency of https://github.com/MesTTo/PeTTaChainer, a PLN reasoner built on top of this runtime, whose README documents a plain pip install ./PeTTa.

I'm filing this as an issue rather than a PR because the obvious fix, py_modules=["petta"] instead of find_packages, only solves half the problem. petta.py locates its Prolog runtime relative to its own file (os.path.dirname(__file__)/../src/main.pl, ../python/helper.pl, ../mork_ffi/...), which assumes the source-tree layout where src/ sits next to python/. In a real non-editable install petta.py ends up inside site-packages with nothing under src/ alongside it, so fixing the module discovery alone wouldn't make a non-editable install work end to end, it would still fail at runtime looking for main.pl. A correct fix needs the Prolog resources moved into the installed package layout (package_data under python/, say) and petta_path resolution updated to match. That's a real repackaging job, not a one-line change, and I'd rather describe it precisely than send a partial fix that looks complete.

Repro:

git clone https://github.com/trueagi-io/PeTTa.git
cd PeTTa
python -m venv .venv && .venv/bin/pip install .
.venv/bin/python -c "import petta"
# ModuleNotFoundError: No module named 'petta'

Tested on Linux (Ubuntu 26.04 LTS, x86_64). Haven't checked macOS or Windows.

Leul-Negash:
## Description

Fixes #190.
`pip install .` produced a wheel with only `tests/` and dist-info, so `import petta` failed with `ModuleNotFoundError`. `petta.py` also resolved its Prolog runtime relative to the source tree, which is absent in a non-editable install.

## Updates
1. Turned `python/petta.py` into the `python/petta` package so the module is discovered and can carry bundled resources.
2. Ship the runtime (`src/`, `lib/`, `python/helper.pl`) inside the wheel via a `build_py` step that copies them into `petta/_runtime/`, preserving the `src/`–`lib/` sibling layout `metta.pl` relies on (`library_path = src/../lib`).
3. Resolve the runtime via `PETTA_PATH`, then the bundled `_runtime/`, then the source-tree root, with a clear error if none is found.
4. Extended `MANIFEST.in` for sdist builds and stopped leaking `tests` as a top-level package.

## Validation
- Non-editable `pip install .`, `PETTA_PATH` override, and `pip install -e .` all run MeTTa end-to-end.
- Existing `python/tests/` suite: 3/3 pass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
 python/__pycache__
 python/HE/__pycache__
 python/tests/__pycache__
+python/petta/__pycache__
+
+# Python build artifacts
+build/
+dist/
+*.egg-info/
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,3 @@
 recursive-include src *
+recursive-include lib *
+include python/helper.pl
--- a/python/petta/__init__.py
+++ b/python/petta/__init__.py
@@ -6,6 +6,25 @@ CONSULTED = False
 CONSULT_LOCK = threading.Lock()
 janus = None
 
+
+def _resolve_petta_path():
+    """Locate the PeTTa runtime tree (src/, lib/, python/helper.pl).
+
+    Prefers PETTA_PATH, then the runtime bundled in the installed package,
+    then the source-tree root (editable installs and checkouts).
+    """
+    env_path = os.environ.get("PETTA_PATH")
+    if env_path:
+        return os.path.abspath(env_path)
+
+    here = os.path.dirname(os.path.abspath(__file__))
+    bundled = os.path.join(here, "_runtime")
+    if os.path.exists(os.path.join(bundled, "src", "main.pl")):
+        return bundled
+
+    return os.path.abspath(os.path.join(here, os.pardir, os.pardir))
+
+
 class PeTTa:
     def __init__(self, verbose=False, petta_path=None):
         global CONSULTED, janus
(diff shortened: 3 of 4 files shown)


Leul-Negash:
@rTreutlein could you take a look when you get a chance? This fixes #190 and touches `setup.py` and `MANIFEST.in`, which you own, and it also changes `python/petta.py`, so @patham9 for visibility since you last touched it. CI is green, and I checked that non-editable `pip install .`, `pip install -e .`, and a `PETTA_PATH` override all run MeTTa end to end.

rTreutlein:
Looks good just need to update the requirement to python >=3.8 (for the shutil.copytree dirs_exist_ok param)

Leul-Negash:
> Looks good just need to update the requirement to python >=3.8 (for the shutil.copytree dirs_exist_ok param)

 @rTreutlein bumped python_requires to >=3.8.

rTreutlein (approved):

PR #193, opened 2026-07-08
fix: allow py-call method calls on strings and numbers (#188)

### LINKED ISSUE #188: Calling methods via py-call on strings throws "ModuleNotFoundError: No module named ..."
I am not sure whether this issue is janus-swi library issue or PeTTa/janus-swi integration issue. But when I try to call methods on strings or numbers using `py-call` the object is interpreted as a module instead.

For example the following code is expected to return "abc".
```metta
!(import! &self ./lib/lib_llm)
!(let $s (py-str ("b")) (py-call (.join $s ("a" "c"))))
```
it shows the following error instead:
```
b
ERROR: [Thread main] /home/vital/projects/trueagi/PeTTa/src/main.pl:23: user:main Python 'ModuleNotFoundError':
ERROR: [Thread main]   No module named 'b'
```

At the same time calling methods on custom objects works:
```metta
!(import! &self ./t.py)
!(let $a (py-call (t.newA)) (py-call (.foo $a)))
```

`t.py`:
```python
class A:
    def foo(self):
        print("A.foo() is called")

def newA():
    return A()

```

Leul-Negash:
Fixes #188.

Calling a method on a string or number via `py-call` raised `ModuleNotFoundError: No module named ...`. For example:

```metta
!(let $s (py-str ("b")) (py-call (.join $s ("a" "c"))))
; before: ERROR ... Python 'ModuleNotFoundError': No module named 'b'
; after:  abc
```

## Cause

The `.method` branch of `py-call` runs `py_call(Obj:Meth)`, and janus reads the left-hand side of `:` as a **module name**. Real object references work because janus keeps them as opaque handles, but converted values (`str`, `int`, …) are plain Prolog terms, so janus tries to import them as modules — hence `No module named 'b'`.

## Fix

Branch on the receiver in the `.method` case (`src/metta.pl`):

- `py_is_object(Obj)` → keep `py_call(Obj:Meth)` (real object references — **unchanged**).
- otherwise → dispatch through the value's type: `py_call(type(Obj):Fun(Obj, Args...))`.

## Before / After

| Call | Before | After |
| --- | --- | --- |
| `(py-call (.upper "abc"))` | `ModuleNotFoundError` | `ABC` |
| `(py-call (.join "-" ("a" "b" "c")))` | `ModuleNotFoundError` | `a-b-c` |
| `(py-call (.replace "abc" "a" "x"))` | `ModuleNotFoundError` | `xbc` |
| `(py-call (.split "a,b,c" ","))` | `ModuleNotFoundError` | `(a b c)` |
| `(py-call (.__add__ 5 3))` | `ModuleNotFoundError` | `8` |

## Validation

- The issue's original example now returns `abc`.
- Existing paths are untouched: object methods (`.foo`, `.bar $a 3 4`), `mod.fun` (`math.sqrt`), and bare `fun` (`str`) all still work.
- `sh test.sh`: all 144 example tests pass. Python `pytest`: 3/3 pass.

### DIFF
--- a/examples/python.metta
+++ b/examples/python.metta
@@ -10,3 +10,7 @@
               ($temp (set-attribute $obj foo (math.pi))))
              (get-attribute $obj foo))
        3.141592653589793)
+
+;Verify py-call on methods of strings and numbers (see #188)
+!(test (py-call (.upper "abc")) ABC)
+!(test (py-call (.__add__ 5 3)) 8)
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -224,10 +224,14 @@ assert(Goal, true) :- ( call(Goal) -> true
                                         ( sub_atom(A, 0, 1, _, '.')         % ".method"
                                           -> sub_atom(A, 1, _, 0, Fun),
                                              Args = [Obj|Rest],
-                                             ( Rest == []
-                                               -> compound_name_arguments(Meth, Fun, [])
-                                                ; Meth =.. [Fun|Rest] ),
-                                             py_call(Obj:Meth, Result, Opts)
+                                             ( py_is_object(Obj)            % on a Python object reference
+                                               -> ( Rest == []
+                                                    -> compound_name_arguments(Meth, Fun, [])
+                                                     ; Meth =.. [Fun|Rest] ),
+                                                  py_call(Obj:Meth, Result, Opts)
+                                                ; py_call(builtins:type(Obj), Ty), % on a converted value (str, int, ...)
+                                                  Call =.. [Fun, Obj|Rest],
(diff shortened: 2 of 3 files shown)


vsbogd (commented):
Thank you for this fix. Looks good to me. But please merge https://github.com/LNAMTY/PeTTa/pull/1 to add unit tests for the change.

Leul-Negash:
> Thank you for this fix. Looks good to me. But please merge [LNAMTY#1](https://github.com/LNAMTY/PeTTa/pull/1) to add unit tests for the change.

@vsbogd  I have merged the pr.

vsbogd (approved):

Issue #196, opened 2026-07-09
No way to import module using relative path


vsbogd:
Working on OmegaClaw plugins API for plugins written in MeTTa, I am trying to implement loading plugin by the absolute path. The issue is that if plugin include more than one file (for example one MeTTa file and one Python file) then it cannot load its modules using relative paths.

Simplified example, let's say we have the following files hierarchy:
`d.metta`:
```metta
!(println! "d.metta is imported")
!(import! &self ./a/b.metta)
```

`./a/b.metta`:
```metta
!(println! "b.metta is imported")
!(import! &self ./c.metta)
!(import! &self ./e.py)
```

`./a/c.metta`:
```metta
!(println! "c.metta is imported")
```

`./a/e.py`:
```py
print("e.py is imported")
```

Then `./run.sh d.metta` prints:
```
"d.metta is imported"
"b.metta is imported"
```
but cannot import `a.metta` or `e.py`.

vsbogd:
@rTreutlein probably should be closed by #195 ?

patham9:
This will also be addressed with the import rework.

PR #195, opened 2026-07-08
Relatives Import , Import Errors , Primport signatures , Various Improvments

rTreutlein:

Issue #181, opened 2026-06-05
Order of imports changes the code behavior


vsbogd:
In the case below `./run.sh a.metta` doesn't print `"foo"` because `c` module is imported before `b`. If the modules are imported in reversed order then `"foo"` is printed.

`i.metta`:
```
!(import! &self c)
!(import! &self b)
```

`c.metta`:
```
(= (bar) (foo))
```

`b.metta`:
```
(= (foo) (println! "foo"))
```

`a.metta`:
```
!(import! &self i)
!(bar)
```

Log:
```
--> metta runnable  -->
!(import! &self i)
-->  prolog goal  --> 
:- findall(A, 'import!'('&self', i, A), _).
^^^^^^^^^^^^^^^^^^^^^^^
--> metta runnable  -->
!(import! &self c)
-->  prolog goal  --> 
:- findall(A, 'import!'('&self', c, A), _).
^^^^^^^^^^^^^^^^^^^^^^^
--> metta function -->
(= (bar) (foo))
--> prolog clause -->
bar([foo]).
^^^^^^^^^^^^^^^^^^^^^^
--> metta runnable  -->
!(import! &self b)
-->  prolog goal  --> 
:- findall(A, 'import!'('&self', b, A), _).
^^^^^^^^^^^^^^^^^^^^^^^
--> metta function -->
(= (foo) (println! "foo"))
--> prolog clause -->
foo(A) :-
    'println!'("foo", A).
^^^^^^^^^^^^^^^^^^^^^^
--> metta runnable  -->
!(bar)
-->  prolog goal  --> 
:- findall(A, bar(A), _).
^^^^^^^^^^^^^^^^^^^^^^^
true
(foo)
```

patham9:
@vsbogd This is expected behavior with current import handling. So far import! is merely a function, not a special macro that gets recursively resolved, and since PeTTa is not interpreting function calls dynamically at runtime either, it is not order-independent. Proposals to make it order-independent already exist as PR (such as https://github.com/trueagi-io/PeTTa/pull/161 ) but such a solution will be for next major release v1.1.0.

Issue #180, opened 2026-06-04
Python module import errors are not printed in chained imports


vsbogd:
When Python module is loaded from MeTTa module which in turn is loaded from MeTTa module then any errors inside imported Python module break the import silently. For example:
`a.metta`:
```
!(import! &self b)
```
`b.metta`:
```
!(import! &self "c.py")
```
`c.py`:
```
raise RuntimeException("error")
```
Then `petta a.metta` doesn't show error while `petta b.metta` shows exception. In both cases the import of the `c.py` fails.

patham9:
Another thing to fix with imports. :)

PR #201, opened 2026-07-20
Overhaul import loading semantics

rTreutlein:
## Summary

Overhaul file imports by combining the strongest parts of the existing import PRs into one coherent loader:

- propagate errors from imported MeTTa and Python files;
- recursively pre-register imported function signatures so import order does not affect compilation;
- resolve nested relative imports from the importing file;
- canonicalize import paths;
- make imports idempotent per target space;
- terminate cyclic imports using explicit `loading` and `loaded` states;
- clear failed import state so a repaired source can be retried;
- raise an explicit `existence_error(source_sink, ...)` for missing files;
- move `git-import!` from `lib_import` into the core loader;
- prepare core Git dependencies before signature discovery, then safely no-op when their normal runnable is encountered;
- resolve library names to the first library root containing the requested source.

The signature prescan remains separate from normal module execution, so imported atoms, clauses, and top-level runnables are still processed normally. `git-import!` is the one core dependency primitive prepared before that scan because it can make later imports resolvable.

This consolidates ideas explored in #157, #159, #161, #171, #189, and #195. It addresses #154, #180, #181, and #196.

## Validation

```
bash test.sh
# passed, including a cold clone in examples/git_import.metta
# and the network-free examples/git_import_core.metta regression

cd python
uv run pytest tests
# 4 passed
```


rTreutlein:
Signature prescanning is causing more trouble than it's worth so i will be removing it and creating a new PR.

PR #202, opened 2026-07-20
Extract miscellaneous runtime improvements from import work

rTreutlein:
## Summary

Extract the non-import changes that were previously bundled into #195 so the import overhaul can be reviewed independently.

This branch contains four separate, pre-existing improvements:

- evaluate function expressions used as the target space of `add-atom` and `remove-atom`;
- add `first`, `second`, `cns`, and `flip` pair helpers to `lib_roman`;
- handle CLI silent flags separately from the input filename;
- raise the embedded SWI-Prolog stack limit for Python API users.

## Validation

```
sh run.sh examples/add_atom_fun_space.metta --silent
sh run.sh examples/lib_roman_pair_helpers.metta --silent
cd python && uv run pytest tests/test_petta.py
```

Both focused MeTTa examples passed and the Python suite reported `3 passed`. `git diff --check` is clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,17 @@ The following projects are cloned and built by build.sh:
 
 Please check out [Extension libraries](https://github.com/trueagi-io/PeTTa/wiki/Extension-libraries) for a set of extension libraries that can be invoked from MeTTa files directly from the git repository.
 
+Git imports retain the legacy URL-only, URL/build-command, and
+URL/build-command/base-directory forms. For a reproducible detached checkout,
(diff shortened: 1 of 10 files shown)


patham9 (commented):
Looks good, but still draft?

rTreutlein:
Ready now. Also just as note regarding the higher stack size limit the current run.sh also set this so this just get's the python path on the same level.

patham9 (approved):
Looks good, thank you!

PR #203, opened 2026-07-21
Overhaul import loading semantics

rTreutlein:
Reworks `import!` around runtime, in-order resolution. Replaces #201 with a cleaner approach after review discussion.

## Changes

- **Relative imports resolve from the importing file.** `load_metta_file` maintains a working-directory stack (exception-safe via `setup_call_cleanup`), so a file imported from another directory finds its own relative MeTTa and Python dependencies. Previously a single global `working_dir` was set once at startup.
- **Import errors surface instead of failing silently.** A missing import target throws `existence_error`; syntax errors and Python import failures in dependencies propagate wrapped with the offending filename. Because a throwing `import!` no longer backtracks, `library/2` now prefers the candidate path whose source file exists across all registered library paths.
- **Imports are idempotent and cycle-safe.** Per-space load state keyed by canonical path: repeated imports are no-ops, mutual imports terminate, and a failed load clears its state so the file can be fixed and re-imported in the same session (covered by a pytest).
- **Git import failures surface.** `git-import!` (unchanged location, `lib_import.pl`) now checks clone/build exit status and throws `process_error` instead of discarding output; repository paths register canonicalized.
- **Warning on use-before-import.** A function registered after its name was already compiled as a plain symbol prints a warning telling the user to move the import or definition above the first use — replacing silent non-reduction, the sharpest edge of the old behavior. Import targets are resolved strictly at runtime, in file order, so runtime-generated or computed import paths keep working.

## Testing

- `test.sh`: all 149 examples pass, no spurious warnings across the corpus (new examples cover relative-import nesting, error surfacing, duplicate/cycle imports, and per-space import identity)
- `python/tests`: 5 pass, including new tests for retry-after-failed-import and the use-before-import warning


patham9:
```
!(git-import! "https://github.com/patham9/mettaclaw.git")
```
currently leads to
```
ERROR: [Thread main] /home/nartech/PeTTa/src/main.pl:34: user:main 'git-import!': Domain error: `git_origin' expected, found `'https://github.com/patham9/mettaclaw'' (expected(https://github.com/patham9/mettaclaw.git))
```

PR #204, opened 2026-07-21
Throw on overapplied function calls

rTreutlein:
Having !(+ 1 2 3) resultin in (partial + (1 2 3)) has been bugging me. This is clearly wrong and not actually a partial application. It's an overapplication that can never run and should instead cause an error.


patham9 (approved):
Looks solid, thank you!

patham9 (approved):
Looks solid, thank you!

PR #207, opened 2026-07-22
Fix library call semantics

### LINKED ISSUE #206: What is library function semantics?
Trying to debug the OmegaClaw-Core imports logic I have realized that I don't understand the intended semantics of the `library` function. `library` function is based on `library_path(Path)` predicate. `library_path` is true for the `<PeTTa root>/lib` directory and for each directory imported via `git-import!` (typically it is `./repos/<reponame>`). There are two clauses of `library`:
```prolog
library(X, Path) :- library_path(Base), atomic_list_concat([Base, '/', X], Path).
library(X, Y, Path) :- library_path(Base), atomic_list_concat([Base, '/../', X, '/', Y], Path).
```
First one tries to construct path using `library_path` predicate argument and the argument of the `library`.
Second uses `library_path` argument and concatenates it with `../X/Y` where `X` and `Y` are `library` arguments.

In OmegaClaw-Core code I see imports like `(import! &self (library OmegaClaw-Core ./src/utils))` which I read in the following way: construct the path which is consisted of the path to the `OmegaClaw-Core` library and `./src/utils`. I expect: (1) only the path of the `OmegaClaw-Core` library is used (2) it is the only path. But instead `library` constructs paths starting from each `library_path`  instance and continuing it by `OmegaClaw-Core/src/utils`. As results there are sevel paths are tried and some of them don't exist. Moreover `import!` used with the Python module adds all of these paths into the `PYTHONPATH` list without checking if there are duplicates.

I would suggest fixing this by changing the `library/2` definition to use only the `<PeTTa root>/lib` path and changing `library/3` definition by adding the check the `library_path` is a path to the `X` (first argument of the `library`). We could add checking for duplicates into `import!` as well. This change reflects my understanding of intended semantics of the `library` function.

vsbogd:
Implementing approach suggested in #206:
Change the `library/2` definition to use only the `<PeTTa root>/lib` path and change `library/3` definition by adding the check the `library_path` is a path to the `X` (first argument of the `library`).

### DIFF
--- a/examples/git_import.metta
+++ b/examples/git_import.metta
@@ -3,7 +3,7 @@
 
 !(git-import! "https://github.com/rTreutlein/test_metta_lib.git")
 
-!(import! &self (library test))
+!(import! &self (library test_metta_lib test))
 
 !(test (mytest 10) 10)
 
--- a/src/metta.pl
+++ b/src/metta.pl
@@ -1,10 +1,10 @@
 %%%%%%%%%% Dependencies %%%%%%%%%%
-library(X, Path) :- library_path(Base), atomic_list_concat([Base, '/', X], Path).
+library(X, Path) :- standard_library_path(Base), atomic_list_concat([Base, '/', X], Path).
 library(X, Y, Path) :- library_path(Base), atom_concat(_, X, Base), atomic_list_concat([Base, '/', Y], Path).
 :- prolog_load_context(directory, Source),
    directory_file_path(Source, '..', Parent),
    directory_file_path(Parent, 'lib', LibPath),
-   asserta(library_path(LibPath)).
+   asserta(standard_library_path(LibPath)).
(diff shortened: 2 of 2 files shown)


patham9:
Totally agree. It caused some troubles to have library be nondet which is indeed problematic @rTreutlein probably wanted lib paths to resolve automatically but better to have it explicit. If Roman agrees too we can have this for v1.1.0 as it will be a breaking change.

rTreutlein (approved):
Don't have anything against this.

patham9 (changes requested):
Just blocked merging for now as we first get v1.0.5 done (almost ready) and this will be a breaking change.

vsbogd:
Thanks!

patham9 (approved):
Approved again, will be the first v1.1.0 PR to be merged now.

PR #208, opened 2026-07-23
Fix compile-time divergence in the specializer

rTreutlein:
The specializer can diverge at compile time. A recursive call that wraps its
higher-order argument, e.g.

```
(= (derive $g) $g)
(= (twice $r $g) ($r ($r $g)))
(= (evolve $r $n $g) (if (== $n 0) $g (evolve (twice $r) (- $n 1) $g)))
!(evolve derive 2 stmt)
```

never finishes translating: every nested specialization builds a strictly
larger key (`partial(twice,[derive])`, then
`partial(twice,[partial(twice,[derive])])`, ...) so the ho_specialization
memo never hits. No type declarations involved; reproduces on current main.

Fix: a function that is already on the in-progress specialization stack is
not re-specialized. An identical key would have been memoized, so a nested
occurrence necessarily has a different, growing key - exactly the divergence
signature. The inner call just compiles as a normal unspecialized call, and
since the set of function symbols reachable from a clause body is finite,
refusing same-function re-entry guarantees termination. Chains of distinct
functions still specialize as before.

Also gates the "Not specialized" fallback print behind the silent flag like
the other translator diagnostics.

The repro above is included as examples/specialize_recursive_wrap.metta (it
now compiles and returns the right value); test.sh green.

https://claude.ai/code/session_018hetjCktvza2Xa7nCopnJx

### DIFF
--- a/examples/specialize_recursive_wrap.metta
+++ b/examples/specialize_recursive_wrap.metta
@@ -1,0 +1,9 @@
+; A recursive call that wraps its higher-order argument used to diverge at
+; COMPILE time: each nested specialization built a bigger key
+; (partial(twice,[partial(twice,[derive])]) ...) and never terminated.
+; Re-specializing a function already being specialized is now refused; the
+; inner call compiles unspecialized and translation terminates.
+(= (derive $g) $g)
+(= (twice $r $g) ($r ($r $g)))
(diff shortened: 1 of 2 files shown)


patham9 (approved):
Makes sense, thanks for the improvement!

PR #209, opened 2026-07-23
Compile-time typechecking

rTreutlein:
# Compile-time typechecking

Adds an optional type system to PeTTa. Declarations are checked during
translation, most checks resolve at compile time, and the type information
feeds code generation.

What you get:

- `(: f (-> Number Number Number))` is checked at every call site and
  definition. Mismatches are compile errors with file and line.
- Undeclared functions get types inferred from their bodies — including
  parameters bound inside destructuring patterns. Inference only adds
  knowledge: it eliminates guards and types call outputs, but never demands
  anything of a caller and never rejects a program that would otherwise run.
- Partially typed code gets one inlined check per unresolved position
  (`number(V) -> true ; ...`) instead of reflection per call.
- `--strict`: every function needs a declared or inferable type and no
  implicit runtime checks may remain. Dynamic values are handled visibly,
  either with `(the Type Expr)` (a single checked cast) or by declaring
  `%Undefined%` outputs.
- A full type language: parametric lists, structural tuples
  `(STV Number Number)`, nominal constructor declarations, union types
  `(| A B)` with pattern narrowing, erased nominal newtypes
  (`(: Proof (Newtype Expression))` — compile-time brands with zero runtime
  representation), and match patterns typed from declared relation schemas.
  Heterogeneous lists widen to the union of their element types instead of
  losing their type. Several `.metta` files can be loaded in order, and
  knowledge arriving late — a declaration or constructor in a later file —
  recompiles exactly the clauses whose verdicts it invalidates.

**Determinism is part of the type.** Three arrows: `-[det]->` (exactly one
result), `-[semidet]->` (zero or one, committed identically), `-[nondet]->`.
A written commitment is validated in **every** mode — body analysis with
positive evidence about callees, non-overlapping heads, a provable-only
exhaustiveness check — and compiles to a commit after head match, so deep
deterministic recursion runs in constant memory (22MB instead of 285MB at
depth 3M in the included example). The builtin determinism table was derived
by reading every predicate's source, and call sites with manifest argument
shapes get sharper verdicts than the (name, arity) worst case. A committed
arrow also enforces **bound arguments at the boundary — exactly where the
determinism proof relied on boundness, and nowhere else**: a `(and $l $r)`
wrapper gets a `nonvar` check per consumed parameter and throws a clear error
on an unbound argument instead of silently enumerating, while a pure data
constructor, det with any arguments, gets none. A parameter the body tests
with `is-var` is exempt — the author handled the unbound case. "Typed implies
bound" holding at the consumed positions is what lets Bool-typed parameters
use the boolean builtins, typed parameters reach `remove-atom`, and
fixed-width tuple parameters count as manifest lists — all deterministically.
The analysis is also flow-sensitive where it pays: behind a
`(== $xs ())` test a semidet head accessor is det, and a function whose every
clause provably yields a proper list (a `collapse`, say) certifies its output
so list builtins are det through the wrapper. `--strict-det` remains the
strictest level: every arrow position of a declaration must state its
cardinality explicitly (`-[det]->`, `-[semidet]->`, `-[nondet]->`), higher-
order parameters included — a plain `->` is rejected there rather than
silently reinterpreted. Higher-order conduits declare an effect variable
instead of overloading: `(: map (-[$v]-> (-[$v]-> $a $b) (List $a) (List $b)))`
is det with a det closure and nondet with a nondet one, and a nondet
instantiation consumed in a det context is a compile error.

**The checker audits itself.** `--oracle` re-emits every statically
discharged certification — clause outputs and call-site arguments — as a
runtime check; `--oracle-det` counts the solutions of every committed call
and throws on a cardinality violation; `--no-det-cut` suppresses the commit
itself. The test suite runs all three over every example, plus a manifest of
known-counterexample programs (some multi-file — several holes exist only
across a load-order boundary) that must fail for exactly the stated reason.
Three adversarial audit rounds against these oracles closed eleven soundness
holes; the recurring defect shape — "unknown" read as "compatible with
everything" — is now a design rule: unknown is not evidence, anywhere.

On real code: we converted the PeTTaChainer inference engine to full strict
typing (including the `--strict-det` explicit-arrow migration), and the
conversion surfaced and fixed real bugs in the chainer itself. It also drove
the checker: dozens of reported gaps (union narrowing through `case`, brands
after control flow, accessor element types, space-op determinism, and more)
are fixed and pinned as examples in this branch.

On the ConceptNet own-pet benchmark, the type-directed execution path
reduced chainer query CPU time by **~23%** against master in a controlled
comparison (identical proof and TV, paired runs). The end-to-end branches
are ~4% apart because two effects unrelated to the checker mask the gain on
this particular query: the typed export's corrected conjunction-introduction
semantics adds 5,545 rules to an And-heavy candidate set, and the two
exports fall on opposite sides of SWI-Prolog's 2^20 clause-index boundary
(~2.8%). The interesting part is *where* the win comes from: the typed
query performs **97.2% fewer `reduce` calls** (640 vs 22,769) — statically
typed calls compile to direct predicate calls instead of interpretation —
while runtime type validation is negligible: 84 shallow guards and zero
slow-path value checks in the hot path. Strict typing pays through
compilation, not checking. This is a controlled result for one
conjunction-heavy workload, not a universal speedup claim.

Microbenchmarks, from the individually benchmarked codegen commits:

| example | base | this branch | |
|---|---|---|---|
| fib | 0.674s | 0.389s | 1.73x |
| fibadd | 0.678s | 0.383s | 1.77x |
| hyperpose_primes | 1.940s | 1.085s | 1.79x |
| holbenchmark | 1.980s | 1.387s | 1.43x |
| he_minimalmetta | 4.232s | 5.143s | 0.82x (slower) |

The gains show up where typed values run through arithmetic, closures or
deterministic recursion in inner loops, and wash out in match/search-dominated
code. The he_minimalmetta row is the honest cost: translation got roughly
20-40% more expensive, which you notice as load time and in code that
re-translates its expressions at runtime. Memoizing runtime translations
would remove that whole class of cost (it would help untyped code just as
much); planned as a follow-up.

Implementation notes: there is
one type store, one compatibility relation (`type_unify/2`), and types travel
on attributed variables — no second bookkeeping path. Errors are thrown
during translation; generated code is never re-scanned, and no subexpression
is translated twice. `cons`, `append` and friends deliberately have no global
`(List $a)` signatures (that would reject legal heterogeneous expressions) —
they are typed contextually instead, constructors and accessors both. Runtime
guards go through `get-type`, so user-defined refinement types keep working.

Testing: the example suite went from 152 to 383 files — 117 of them `fail_*`
cases that must be rejected with the exact expected error, the rest running
green, many under `--strict`/`--strict-det` — plus five matrix scripts
(oracle soundness, type dispatch, library trust, builtin-registry
consistency, and load-order independence of the checker units) asserting
properties of the generated code and of the checker itself. `test.sh` is
green after every commit, and the history is granular on purpose: each
codegen optimization is its own commit and was benchmarked individually.

Size: the checker is 14 ownership units under `src/typecheck/` (~5300
lines: one declarative builtin registry, one canonical declaration store,
analyses that return proof records, and one dependency graph that
recompiles exactly the clauses a late declaration or clause change
invalidates); `src/translator.pl` grew to ~1400. The rest of the diff is
tests and README documentation.

https://claude.ai/code/session_018hetjCktvza2Xa7nCopnJx


rTreutlein:
Added since opening: `--strict-det`, a level above `--strict`.

Under it a plain `->` is itself a determinism commitment: every declared
function is validated (no overlapping clause heads, no superpose/match/eval
in the body) unless its arrow says `-[nondet]->`. Since in MeTTa every
matching clause fires, each rejection is either an accidental source of
multiple results or a missing nondet annotation. Closure parameters carry
the same commitment - a `(-> $a $b)` param may be applied in a det body, a
`-[nondet]->` one may not - and a clause that commits with `(cut)` may
overlap later clauses. Functions added at runtime via add-atom go through
the same validation.

The bundled libraries load clean under the flag (the genuinely
nondeterministic helpers are now declared `-[nondet]->`). Implementing this
also surfaced a pre-existing bug in the determinism analyzer: variable
headed body expressions unified with the construct patterns it matches
against, so any var-head body was misclassified - fixed, along with
treating case/let patterns as patterns rather than expressions.

Without the flag nothing changes; `--strict` behaves exactly as before.
Suite is at 207 examples.

rTreutlein:
Two additions since the last update:

**Erased nominal newtypes** (`a3432f6`). Real-world usage (converting the
PeTTaChainer inference engine to strict typing) surfaced a need for semantic
roles - KB, Proof, Statement - that share one runtime representation. Nominal
constructors would change the representation, and value declarations can only
assign one global role per atom, so:

```metta
(: KB    (Newtype Expression))
(: Proof (Newtype Expression))
(: rule-ev (-> KB Proof Evidence))
```

declares distinct compile-time brands that are fully erased: no wrappers, no
guards. A brand fits its representation, but different brands never unify
merely because their representations do - swapping a Proof into a KB position
is a compile error. Raw literals and constructed values acquire a brand
contextually from the expected position; unknown variables do not - they need
the explicit erased trust operation `(brand KB $x)`, which rejects
conflicting brands but generates no check (a role has no runtime witness by
construction; `(the KB ...)` remains the checked alternative for validating
the representation). Declared relation schemas restore brands after untyped
space round-trips. The positive spec example runs under `--strict`, which is
itself the proof that branding emits no residual checks.

**Soundness fixes from an external review** (`663de5c`). A control-flow
review of the checker found nine issues; eight were confirmed by executable
repro and all are fixed and pinned as examples - notably: declared output
type variables now require a genuinely parametric implementation, manual
call/reduce dispatch applies input checks, deterministic functions require
positive determinism evidence about undeclared callees (transitive body
analysis), open structured types guard their shape, and construct output
typing (collapse/foldall/quote/and-then) no longer certifies types the
runtime result need not satisfy.

The branch also merged current main (import overhaul, overapplication
errors), so the PR is mergeable again. Suite is at 222 examples.

https://claude.ai/code/session_018hetjCktvza2Xa7nCopnJx

rTreutlein:
Since the last update the branch went through a full line-by-line review plus
three rounds of adversarial soundness auditing, and a tight feedback loop
with the PeTTaChainer strict-typing conversion. The PR description is
rewritten to describe the current state; the delta since the 23rd:

**Soundness.** Eleven holes closed, found by new self-audit instruments:
`--oracle` now re-verifies call-site *argument* certifications (not just
outputs), and the new `--oracle-det` counts the solutions of every committed
call — the only oracle that can see determinism-model errors. Fixed among
others: branch merges discharged by a single typed branch, `(Newtype
<wildcard>)` acting as a universal type, promised type variables read as
knowledge, 18 wrong entries in the builtin determinism table, `car-atom`
answering `()` for an empty expression where its certified element type was
proven, and compound-headed applications counted as manifest list spines.

**Determinism model.** `-[semidet]->` (zero-or-one, committed) joined the
arrows; every written commitment is now checked in every mode, exhaustiveness
included. Builtin verdicts are argument-aware — a manifest list spine or
bound boolean strengthens the (name, arity) worst case. And a committed
arrow now enforces **bound arguments at the boundary**: an unbound argument
throws a clear error instead of silently enumerating, which is what makes
`and`/`or`/`not` over Bool-typed parameters, `remove-atom` over typed
parameters, and fixed-width tuples as manifest lists all soundly `det`.

**PeTTaChainer results.** The typed chainer runs **15% faster** than the
untyped one — 95% of `reduce` calls avoided, plus other compilation
optimizations the type information enables — and the conversion surfaced and
fixed various real bugs in the chainer itself. Twelve checker gaps it
reported (union narrowing through `case`, brands after control flow,
accessor element types, heterogeneous-list widening, space-op determinism,
...) are fixed and pinned as examples.

**Late knowledge.** Several `.metta` files load in order; a declaration or
constructor arriving in a later file recompiles exactly the clauses whose
verdicts it invalidates — previously it was believed and never enforced.

Suite is at 307 examples (87 expected-fail) plus the oracle matrices, ~2
minutes wall clock in a bounded worker pool.

https://claude.ai/code/session_018hetjCktvza2Xa7nCopnJx

ngeiswei:
I wonder if we could adopt more compact notations such as

* Deterministic arrow type (instead of `-[det]->`)
   ```
   ->
   ```
* Non-deterministic arrow type (instead of `->`)
   ```
   -<
   ```
* Partial deterministic arrow type (instead of `-[semidet]->`)
   ```
   ->p
   ```
* Partial non-deterministic arrow type (new)
   ```
   -<p
   ```

You may notice the new arrow type `-<p` indicating a non deterministic function that is not total.

I understand that it breaks backward compatibility, cause the regular arrow type `->` would then be restricted to deterministic functions, but one can control with flags whether it is ignored, warned or halted by errors, so I don't think it is such a big deal.

What do you think?

rTreutlein:
@ngeiswei I would not replace the standard arrow though i guess we can if we don't care about backwards compatibility. @patham9 opinion?

PR #212, opened 2026-07-24
`Expression -> Atom` as the signal to not evaluate (new PR)

patham9:
Please see https://github.com/trueagi-io/PeTTa/pull/144


pisaev1 (approved):
Looks good

PR #218, opened 2026-08-25
fix(parser): preserve variable identity across swrite GC

zariuq:
We found a strange bug in shared variables becoming decoupled when working on CeTTa --lang petta on Nil's chaining bechmarks.  This var_id style fix seems to work and possibly improve performance.  Feel free to edit the code however you please.

---

Codex Sol: 
 
 ## Problem

  `swrite/2` generated variable names using `term_to_atom/2`. These names
  reflect movable SWI-Prolog stack locations.

  If garbage collection occurred during serialization:

  - one logical variable could print under two names, splitting sharing;
  - distinct variables could print under one name, fabricating sharing.

  Reparsing the output then changes its semantics.

  Minimal witness shape:

  ```text
  (: ax1 (-> $X (-> <padding> $X)))

  Upstream printed the two occurrences differently:

  (: ax1 (-> $_9240 (-> <padding> $_262)))

  This change preserves their identity:

  (: ax1 (-> $_0 (-> <padding> $_0)))

  The MeTTa regression example demonstrates the behavioral difference:

  Upstream:
  is distinct-preserved, should distinct-preserved. ✅
  is split-detected, should sharing-preserved. ❌

  Fixed:
  is distinct-preserved, should distinct-preserved. ✅
  is sharing-preserved, should sharing-preserved. ✅

  ## Fix

  Variable names are minted from a per-call counter and cached temporarily
  on each Prolog variable as an attribute. Attributes follow logical
  variables across moving garbage collection.

  Cleanup removes the temporary attribute without disturbing unrelated
  attributes. The generated naming is deterministic, but the tests depend
  only on preservation of sharing and distinctness.

  ## Verification

  - focused serializer tests: 7/7 pass;
  - full PeTTa example suite passes;
  - Nil OBC S13 returns all 35,063 ordered answers with correct sharing;
  - large distinct-variable stress test produces no name collisions.

### DIFF
--- a/examples/swrite_variable_identity.metta
+++ b/examples/swrite_variable_identity.metta
@@ -1,0 +1,30 @@
+;; Serialized variable identity, asserted behaviorally -- no naming scheme
+;; is assumed anywhere in this file.
+;; (Compact form of the sharing shape from Nil's OBC S13 enumeration.)
+
+(= (countdown $n)
+   (if (< 0 $n)
+       (cons $n (countdown (- $n 1)))
+       ()))
+
+(= (padded-ax1 $p $q)
+   (: ax1 (-> $p (-> $q $p))))
+
+(= (two-vars $p $q)
+   (: pair (-> $p $q)))
+
+;; 1. Distinctness survives the round trip: two distinct variables must
+;;    re-parse as distinct, i.e. bindable to different constants.
+!(test (case (parse (repr (two-vars $a $b)))
+             (((: pair (-> 7 8)) distinct-preserved)
+              ($other fabricated-sharing)))
+       distinct-preserved)
+
+;; 2. Sharing survives the round trip: if the two occurrences of one
+;;    variable print under different names, the re-parsed term has two
+;;    independent variables, which CAN bind to the different constants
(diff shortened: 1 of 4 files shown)


patham9:
I don't get it yet. The example `(: ax1 (-> X (-> <padding> X)))` include no variable at all.
If they would be variables ($X) both occurrences would have the same ID. Now would the same var ID be preserved after swrite and follow-up sread? No, also does not need to. I don't see the issue here.

patham9 (approved):
ah I see the issue is when printing terms, understood now. Good catch, thank you!

zariuq:
Sorry for the sloppy paste w/o $X.  Yeah, it's kinda weird to abuse via parse and repr.  It's great to have 2+ implementations of a language as that makes spotting bugs easier  🥳

```
  (= (countdown $n) (if (< 0 $n) (cons $n (countdown (- $n 1))) ()))
  (= (padded-ax₁ $𝜑 $𝜓) (: ax₁ (→ $𝜑 (→ $𝜓 $𝜑))))
  !(padded-ax₁ $shared (countdown 2000))
```

